PR #236, opened 2025-01-12
unable to do shorts

7ayd:
All short positions are being submitted as longs. Currently unable to submit shorts.


packages/app/src/lib/components/foil/transactionTable.tsx
```diff
@@ -139,7 +139,7 @@ const TransactionTable: React.FC<Props> = ({ transactions }) => {
       );
     }
 
-    if (['Collateral', 'Ggas', 'wstETH', 'Price'].includes(cell.column.id)) {
+    if (['collateral', 'ggas', 'wsteth', 'price'].includes(cell.column.id)) {
```
noahlitvin, line 142: Not sure we want this change here?

noahlitvin (changes requested):

PR #242, opened 2025-01-15
OLD - FOIL 132: Replace Uni quoter "Unexpected error" for a better text

leomassazza:


leomassazza:
Closed in favor of https://github.com/foilxyz/foil/pull/257 which has verified commits

PR #254, opened 2025-01-19
(Redone) Fix settle dust: Reduce deposited collateral after withdrawal

leomassazza:


noahlitvin (approved):

packages/protocol/src/market/storage/Position.sol
```diff
@@ -238,7 +238,8 @@ library Position {
 
         rebalanceEthTokens(self);
         rebalanceCollateral(self);
-
+        // after rebalancing the Eth Tokens and Collateral, all virtual tokens are zeroed and only the deposited collateral is left with the net balance.
```
noahlitvin, line 241:
```suggestion
        // after rebalancing the Eth Tokens and Collateral, all virtual tokens are zeroed and only the deposited collateral is left with the next balance.
```

PR #282, opened 2025-01-30
📈 chart toggles etc

noahlitvin:


packages/api/src/routes/resources.ts
```diff
@@ -1,8 +1,11 @@
 import { Router } from 'express';
 import { handleAsyncErrors } from '../helpers/handleAsyncErrors';
+import { justifyTimeSeries } from '../helpers/timeSeriesHelpers';
```
leomassazza, line 3: do we need the new imports here?

leomassazza (approved):

PR #296, opened 2025-02-06
Get trailing avg from graphql api

leomassazza:


packages/api/src/graphql/resolvers/CandleResolver.ts
```diff
@@ -72,6 +79,93 @@ const groupPricesByInterval = (
...
+  // Normalize timestamps to interval boundaries
+  const normalizedStartTimestamp =
+    Math.floor(startTimestamp / intervalSeconds) * intervalSeconds;
+  const normalizedEndTimestamp =
+    Math.floor(endTimestamp / intervalSeconds) * intervalSeconds;
+
+  // Initialize lastClose with lastKnownPrice if available, otherwise use first price
+  let lastClose = lastKnownPrice || prices[0].value;
+
+  const orderedPrices = prices.sort((a, b) => a.timestamp - b.timestamp);
```
ukitta555, line 104: aren't they already sorted?

packages/api/src/graphql/resolvers/CandleResolver.ts
```diff
@@ -72,6 +79,93 @@ const groupPricesByInterval = (
...
+      endIdx--;
+    }
+
+    // If not found, use the last index of the orderedPrices array
+    if (endIdx == -1) {
+      endIdx = orderedPrices.length - 1;
+    }
+
+    // Add to the sliding window trailing average the prices that are now in the interval
+    for (let i = lastEndIdx; i <= endIdx; i++) {
```
ukitta555, line 144: we could try to make the calculations faster; right now the runtime is approx. O(n^2) since we have a nested loop. If the performance is not too big of a concern we should be good though

packages/app/src/lib/hooks/useChart.ts
```diff
@@ -306,6 +327,53 @@ export const useChart = ({
...
+        resourceSlug,
+        market?.epochId,
+        selectedInterval,
+      ],
+      queryFn: async () => {
+        if (!resourceSlug) {
+          return [];
+        }
+        const now = Math.floor(Date.now() / 1000);
+        const from = now - 28 * 24 * 60 * 60; // Two periods ago
```
ukitta555, line 343: times 2?

ukitta555 (approved):

PR #311, opened 2025-02-13
Fix race condition in indexer

7ayd:
Removed InsertAfter hook for Events and added  upsertEntitiesFromEvent to upsertEntites in market.ts. This should solve the race condition when adding transactions to the transaction table after an event.


packages/api/src/controllers/market.ts
```diff
@@ -450,32 +450,64 @@ const upsertEvent = async (
...
+  try {
+    // Check if event already exists
+    const existingEvent = await eventRepository.findOne({
+      where: {
+        transactionHash: logData.transactionHash,
+        market: { id: market.id },
+        blockNumber: Number(blockNumber),
+        logIndex: logIndex,
+      },
+      relations: ['market', 'market.marketParams'],
```
noahlitvin, line 474: Is `'market.marketParams'` supposed to be here?

7ayd: I adjusted it. I didn't realize that was getting loaded before right above it.

noahlitvin (approved):

PR #313, opened 2025-02-13
FOIL-231

noahlitvin:


packages/api/src/graphql/resolvers/ResourceResolver.ts
```diff
@@ -24,7 +27,7 @@ export class ResourceResolver {
     try {
       const resource = await dataSource.getRepository(Resource).findOne({
         where: { slug },
-        relations: ['markets', 'resourcePrices'],
+        relations: ['markets', 'markets.epochs', 'resourcePrices'],
```
noahlitvin, line 30: might not need this change here

packages/api/src/models/Market.ts
```diff
@@ -35,6 +35,12 @@ export class Market {
   @Column({ type: 'varchar' })
   address: string;
 
+  @Column({ type: 'varchar' })
+  vaultAddress: string;
+
+  @Column({ type: 'boolean', default: false })
+  isYin: boolean;
```
leomassazza, line 42: np: This fixes the architecture for only 2 markets. An id to identify the order would be better (vaults support N markets)

noahlitvin: Good point, I suggest we keep as-is here for now for simplicity, can easily migrate when needed

leomassazza (approved):

PR #323, opened 2025-02-17
Fix initial value set to something different than zero

llamalm:


packages/api/src/graphql/resolvers/CandleResolver.ts
```diff
@@ -39,7 +39,7 @@ const groupPricesByInterval = (
     Math.floor(endTimestamp / intervalSeconds) * intervalSeconds;
 
   // Initialize lastClose with lastKnownPrice if available, otherwise use first price
-  let lastClose = lastKnownPrice || prices[0].value;
+  let lastClose = lastKnownPrice ;//|| prices[0].value;
```
noahlitvin, line 42:
```suggestion
  let lastClose = lastKnownPrice;
```

noahlitvin (changes requested):

noahlitvin (approved):

PR #342, opened 2025-03-03
Added which resource to refresh on`/refresh` endpoint

7ayd:
Handles accepting refreshing specific resources to the /refresh endpoint. Will hard refresh if "hardInitialize" is set to true or else it will soft refresh.


noahlitvin:
added a ui for this in `/admin` so you'll want to update that as well I think

PR #386, opened 2025-03-28
add tia to mainnet

sunnyvempati:


packages/api/src/fixtures.ts
```diff
@@ -125,14 +125,14 @@ const addGasWeekly = async (markets: MarketInfo[]) => {
 
 const addTiaBlobCumulative = async (markets: MarketInfo[]) => {
   const tiaBlobCumulative = await safeRequire(
-    `@/protocol/deployments/outputs/${sepolia.id}-tia-blobs-cum-03-18/Foil.json`
+    `@/protocol/deployments/outputs/${base.id}-tia-blobs-cumulative/Foil.json`
```
sunnyvempati, line 128: just changed this since we don't need that other deployment anymore.  can prolly just live in the db regardless

noahlitvin (approved):

PR #390, opened 2025-03-31
Create audit_reports.mdx

Rafanator:


packages/docs/docs/pages/audit_reports.mdx
```diff
@@ -0,0 +1,5 @@
+## Audit Reports
+
+- [1st Audit – September 30th, 2024](https://raw.githubusercontent.com/foilxyz/foil/main/packages/docs/docs/public/images/2024-09-30_Foil.pdf)
```
noahlitvin, line 3: I think this should be a relative path like `/images/2024-09-30_Foil.pdf`

noahlitvin (approved):

PR #404, opened 2025-04-08
Add price to quoteTrade 

llamalm:


packages/protocol/src/market/interfaces/ITradeModule.sol
```diff
@@ -37,11 +37,12 @@ interface ITradeModule {
      * @param size The position size.
      * @return requiredCollateral The required collateral.
      * @return fillPrice The virtual tokens trade fill price.
+     * @return price18DigitsAfter The price after the trade.
      */
     function quoteCreateTraderPosition(
         uint256 epochId,
         int256 size
-    ) external returns (uint256 requiredCollateral, uint256 fillPrice);
+    ) external returns (uint256 requiredCollateral, uint256 fillPrice, uint256 price18DigitsAfter);
```
sunnyvempati, line 45: can we rename to `priceAfterD18`

sunnyvempati (approved):

PR #410, opened 2025-04-09
Updating Market model

7ayd:
Adding question field to market model and migration.

### DIFF
--- a/packages/api/src/fixtures.json
+++ b/packages/api/src/fixtures.json
@@ -65,7 +65,7 @@
         "What will be the average gas cost on Ethereum on the week of April 1st?",
         "What will be the average gas cost on Ethereum on the week of April 7th?"
       ],
-      "question": "What will be the average gas cost on Ethereum on the week of April 7th?",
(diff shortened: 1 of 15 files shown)


packages/api/src/fixtures.json
```diff
@@ -52,6 +53,7 @@
       "questions": [
         "What will be the average blob gas cost on Ethereum in April?"
       ],
+      "question": "What will be the average blob gas cost on Ethereum in April?",
```
noahlitvin, line 56:
```suggestion
      "question": "What will be the average blob gas cost on Ethereum in a given month?",
```

packages/api/src/fixtures.json
```diff
@@ -42,6 +42,7 @@
   "MARKETS": [
     {
       "questions": ["How much blob data will be posted to Celestia in April?"],
+      "question": "How much blob data will be posted to Celestia in April?",
```
noahlitvin, line 45:
```suggestion
      "question": "How much blob data will be posted to Celestia in a given month?",
```

PR #456, opened 2025-04-23
Add epoch claim statement as override

llamalm:


packages/protocol/src/vault/Vault.sol
```diff
@@ -291,13 +300,25 @@ contract Vault is IVault, ERC20, ERC165, ReentrancyGuardUpgradeable {
...
 
+        // Build the claim statement for the new epoch
+        bytes memory claim = abi.encodePacked(
+            string(claimStatement),
+            " between timestamps ",
+            Strings.toString(startTime),
+            " and ",
+            Strings.toString(startTime + duration),
+            " (inclusive) is "
+        );
```
sunnyvempati, line 311: can we move the start time and end time stuff to the UMASettlementModule and let it append this data?  just weird to have the statement end with "is".  would be nice to have like, and the interpolation can all live in one place instead of some here and some in uma module:

`${claimStatement}, from: ${startTime} to: ${endTime} with price: {decimalPrice}`

sunnyvempati (approved):
looks good, have a suggestion for keeping interpolation in one place to make things cleaner.

PR #507, opened 2025-05-12
Percentage fix

7ayd:
Screenshot 2025-05-11 at 9 58 52 PM

### DIFF
--- a/packages/sapience/src/app/forecasting/[chainShortName]/page.tsx
+++ b/packages/sapience/src/app/forecasting/[chainShortName]/page.tsx
@@ -118,14 +118,20 @@ const ForecastingDetailPage = () => {
   const optionNames = marketGroupData.markets.map(
     (market) => market.optionName || ''
   );
-
   // Otherwise show the main content
   return (
     <div className="flex flex-col w-full min-h-[100dvh] overflow-y-auto lg:overflow-hidden pt-28 pb-40 lg:pt-32 lg:pb-12">
       <div className="container mx-auto max-w-5xl flex flex-col">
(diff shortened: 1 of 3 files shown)


packages/sapience/src/app/forecasting/[chainShortName]/page.tsx
Comment (7ayd) [inline packages/sapience/src/app/forecasting/[chainShortName]/page.tsx:1]: Just incase the marketgroup has multple active markets it pull the question from the market that has a matching address

Comment (noahlitvin) [reply packages/sapience/src/app/forecasting/[chainShortName]/page.tsx:1]: Great! Will merge as-is but let's consider moving this to a util so it's shared between this and the parent page.

packages/sapience/src/components/forecasting/MarketGroupChart.tsx
```diff
@@ -58,7 +58,20 @@ const MarketGroupChart: React.FC<MarketGroupChartProps> = ({
         typeof point.indexClose === 'number'
           ? point.indexClose / 1e18 // Scale Wei down by 10^18
           : point.indexClose; // Keep null/undefined as is
-      return { ...point, indexClose: scaledIndexClose };
+
+      const scaledMarkets: { [marketId: string]: number | undefined } = {};
+      if (point.markets) {
+        Object.entries(point.markets).forEach(([marketId, value]) => {
+          scaledMarkets[marketId] =
+            typeof value === 'number' ? value / 1e18 : value;
```
noahlitvin, line 66: This case could just be for when it's undefined, might be a clearer way to write this?

noahlitvin (approved):

PR #515, opened 2025-05-15
wallet balance

noahlitvin:


packages/sapience/src/components/forecasting/forms/inputs/WagerInput.tsx
```diff
@@ -39,23 +77,12 @@ export function WagerInput({
...
-          className="text-primary underline"
-        >
-          Get sUSDS
-        </a>
-      </>
-    ) : undefined;
+  const collateralSymbol = marketGroupData?.collateralSymbol || '';
+  const collateralAddress = marketGroupData?.collateralAsset as
+    | `0x${string}`
+    | undefined;
```
sunnyvempati, line 85: instead of using the useMarketGroupPage, pass this data in.  This way wecan re-use WagerInput isn't context dependant (for when we move the component into the `ui` directory for example)

sunnyvempati (approved):
pre-approving but would be good to consider the comment

PR #556, opened 2025-06-13
Close converted position after closing LP

llamalm:


packages/protocol/src/market/modules/LiquidityModule.sol
```diff
@@ -247,6 +249,74 @@ contract LiquidityModule is ReentrancyGuardUpgradeable, ILiquidityModule {
...
+            uint256 decreasedAmount0,
+            uint256 decreasedAmount1,
+            uint256 collateralAmount
+        )
+    {
+        if (params.deadline < block.timestamp) {
+            revert Errors.TransactionExpired(params.deadline, block.timestamp);
+        }
+
+        if (params.liquiditySlippage > 1e18 || params.tradeSlippage > 1e18) {
```
sunnyvempati, line 268: we have `DecimalMath.UNIT` that you can use instead

packages/protocol/src/market/modules/LiquidityModule.sol
```diff
@@ -643,6 +715,108 @@ contract LiquidityModule is ReentrancyGuardUpgradeable, ILiquidityModule {
                 deltaCollateral: deltaCollateral
             })
         );
+
+        // Notice: closing the trade position after the event is emitted to have both events show the valid intermediate state
```
sunnyvempati, line 719: yeah makes sense

packages/protocol/src/market/modules/LiquidityModule.sol
```diff
@@ -643,6 +715,108 @@ contract LiquidityModule is ReentrancyGuardUpgradeable, ILiquidityModule {
...
+        int256 deltaCollateralLimit = -int256(
+            position.depositedCollateralAmount.mulDecimal(1e18 - tradeSlippage)
+        );
+
+        Trade.QuoteOrTradeInputParams memory inputParams = Trade
+            .QuoteOrTradeInputParams({
+                oldPosition: position,
+                initialSize: position.positionSize(),
+                targetSize: 0,
+                deltaSize: -position.positionSize(),
```
sunnyvempati, line 740: seems odd we need both these, couldn't you derive one from the other?

packages/protocol/src/market/modules/LiquidityModule.sol
```diff
@@ -643,6 +715,108 @@ contract LiquidityModule is ReentrancyGuardUpgradeable, ILiquidityModule {
...
+            inputParams
+        );
+
+        position.vEthAmount = outputParams.position.vEthAmount;
+        position.vGasAmount = outputParams.position.vGasAmount;
+        position.borrowedVEth = outputParams.position.borrowedVEth;
+        position.borrowedVGas = outputParams.position.borrowedVGas;
+        position.depositedCollateralAmount = outputParams
+            .position
+            .depositedCollateralAmount;
```
sunnyvempati, line 755: maybe a `position.update(..)` function?

packages/protocol/src/market/modules/LiquidityModule.sol
```diff
@@ -643,6 +715,108 @@ contract LiquidityModule is ReentrancyGuardUpgradeable, ILiquidityModule {
...
+        if (position.vGasAmount > 0) {
+            // Notice. This error should not happen. If it's here it means something went wrong
+            revert Errors.InvalidData("Cannot close position with vGas tokens");
+        }
+        if (position.borrowedVGas > 0) {
+            // Notice. This error should not happen. If it's here it means something went wrong
+            revert Errors.InvalidData(
+                "Cannot close position with borrowed vGas tokens"
+            );
+        }
```
sunnyvempati, line 770: these checks seem unnecessary since the code paths leading to this code ensure this can't happen but 🤷🏽

packages/protocol/src/market/modules/LiquidityModule.sol
```diff
@@ -643,6 +715,108 @@ contract LiquidityModule is ReentrancyGuardUpgradeable, ILiquidityModule {
...
+
+        // 2. Confirm collateral is enough to pay for borrowed veth
+        if (
+            position.borrowedVEth > 0 &&
+            position.borrowedVEth > position.depositedCollateralAmount
+        ) {
```
sunnyvempati, line 776: we don't need this check if we're checking `borrowedVEth > 0` above this check

packages/protocol/src/market/modules/LiquidityModule.sol
```diff
@@ -643,6 +715,108 @@ contract LiquidityModule is ReentrancyGuardUpgradeable, ILiquidityModule {
...
+
+        // 5. Set the position kind to unknown
+        position.kind = IFoilStructs.PositionKind.Unknown;
+
+        // Check if the collateral is within the limit
+        Trade.checkDeltaCollateralLimit(deltaCollateral, deltaCollateralLimit);
+
+        uint256 finalPrice = epoch.getReferencePrice();
+
+        epoch.validateCurrentPoolPriceInRange();
```
sunnyvempati, line 801: maybe move this up to right after the trade

sunnyvempati (changes requested):
i think we can simplify all the checks after the trade is completed when closing from LP position.  see if we can, it doesn't hurt to have them so i think can keep some

PR #558, opened 2025-06-18
Use LZ to bridge UMA claims

llamalm:


packages/protocol/src/bridge/MarketLayerZeroBridge.sol
```diff
@@ -0,0 +1,283 @@
...
+        );
+
+        // Check if the asserter has enough bond
+        require(
+            remoteSubmitterBalances[asserter][currency] >=
+                bond + remoteSubmitterWithdrawalIntent[asserter][currency],
+            "Asserter does not have enough bond"
+        );
+
+        // TODO: check if we need to verify other stuff here
```
sunnyvempati, line 247: .

packages/protocol/src/bridge/UMALayerZeroBridge.sol
```diff
@@ -0,0 +1,326 @@
...
+                optimisticOracleV3Address
+            );
+
+        IERC20 bondToken = IERC20(bondTokenAddress);
+
+        bondToken.approve(address(optimisticOracleV3), bondAmount);
+
+        bytes32 umaAssertionId = optimisticOracleV3.assertTruth(
+            claim,
+            asserter,
```
sunnyvempati, line 261: shouldn't the bond be returned to this contract instead of the asserter?  and the asserter here is an eoa on converge?

packages/protocol/remappings.txt
```diff
@@ -2,4 +2,6 @@
 @uma/=node_modules/@uma
 @openzeppelin/=node_modules/@openzeppelin
 @synthetixio/=node_modules/@synthetixio
+@layerzerolabs/=node_modules/@layerzerolabs
+solidity-bytes-utils/contracts/=node_modules/@layerzerolabs/toolbox-foundry/lib/solidity-bytes-utils
```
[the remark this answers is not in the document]
leomassazza: Confirmed. This is used in a lib include and needs the contracts in the remappings path.

PR #636, opened 2025-07-30
feat(frontend): SSR/SEO to forecast page

nicosampler:
In this PR, SSR prefetching of data used on the forecast page was added.

The flow is pretty simple: on the server, we fetch the API to get the data that will be requested by the hooks within ForecastPage. Once the data is returned, we hydrate useQuery, so when the client renders for the first time, the cache is already populated.

### DIFF
--- a/packages/sapience/src/app/forecast/page.tsx
+++ b/packages/sapience/src/app/forecast/page.tsx
@@ -1,301 +1,35 @@
-'use client';
+import { QueryClient, dehydrate } from '@tanstack/react-query';
+import { prefetchEnrichedMarketGroups } from '~/hooks/graphql/useMarketGroups';
+import ForecastPageImp from '~/app/forecast/ForecastPageImp';
+import Hydrate from '~/components/Hydrate';
+import { CONVERGE_SCHEMA_UID } from '~/lib/constants/eas';
+import { prefetchPredictions } from '~/hooks/graphql/usePredictions';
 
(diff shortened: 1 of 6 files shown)


packages/sapience/src/app/forecast/page.tsx
```diff
@@ -1,301 +1,35 @@
...
-                >
-                  <div className="rounded-full p-0.5 w-4 h-4 flex items-center justify-center bg-zinc-500/20">
-                    <FileTextIcon className="w-2.5 h-2.5 text-zinc-500" />
-                  </div>
-                  <span className="font-medium">Selected Question</span>
-                </button>
-              )}
+  // Prefetch enriched market groups data
+  await prefetchEnrichedMarketGroups(serverQC);
+  await prefetchPredictions(serverQC, CONVERGE_SCHEMA_UID);
```
ukitta555, line 26: should this be in getServerSideProps()?

https://tanstack.com/query/latest/docs/framework/react/guides/ssr
https://nextjs.org/docs/pages/api-reference/functions/get-server-side-props

ukitta555 (approved):

packages/sapience/src/app/forecast/page.tsx
nicosampler: You are viewing the router documentation pages. This projects use App Router.

PR #639, opened 2025-07-31
Markets-seo

nicosampler:


packages/sapience/src/components/forecasting/MarketGroupsList.tsx
```diff
@@ -687,19 +689,27 @@ const ForecastingTable = () => {
             {groupedMarketGroups.length > 0 && (
               <motion.div
                 key="results-container"
-                initial={{ opacity: 0 }}
-                animate={{ opacity: 1 }}
-                exit={{ opacity: 0 }}
-                transition={{ duration: 0.25 }}
+                {...(isClient
```
nicosampler, line 692: These animations on SSR prevent the generation of static content.

PR #640, opened 2025-07-31
fix L-20

llamalm:


sunnyvempati (changes requested):
i dont think we should add this complexity, the function is very hard to follow tbh.  plz ignore this as i dont see how this can even happen when we check collateral requirements.

PR #647, opened 2025-08-06
feat(app): batch wagers

nicosampler:


packages/sapience/src/components/forecasting/forms/inputs/MultipleChoiceWager.tsx
```diff
@@ -13,15 +13,17 @@ interface MultipleChoicePredictProps {
   name?: string;
   options: Array<{ name: string; marketId: number }>;
   variant?: 'buttons' | 'dropdown';
+  defaultValue?: string;
```
nicosampler, line 16: used to initialize the correct option when the prediction pop-up is open.

PR #672, opened 2025-08-21
Audit remediation fixes - Highs and misc

llamalm:
H-01
H-02
L-04

### DIFF
--- a/packages/protocol/src/bridge/abstract/BondManagement.sol
+++ b/packages/protocol/src/bridge/abstract/BondManagement.sol
@@ -41,11 +41,15 @@ abstract contract BondManagement is ReentrancyGuard, IBondManagement {
         address bondToken,
         uint256 amount
     ) external virtual nonReentrant returns (MessagingReceipt memory) {
-        require(
-            amount >= _getMinimumDepositAmount(),
-            "Amount must be greater than minimum deposit amount"
-        );
-        require(_isValidToken(bondToken), "Invalid bond token");
+        if (bondToken == address(0)) {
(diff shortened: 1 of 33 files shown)


packages/protocol/src/bridge/abstract/BondManagement.sol
```diff
@@ -83,6 +83,12 @@ abstract contract BondManagement is ReentrancyGuard, IBondManagement {
             withdrawalIntents[msg.sender][bondToken].amount == 0,
             "Withdrawal intent already exists"
         );
+        require(
+            block.timestamp >=
+                withdrawalIntents[msg.sender][bondToken].timestamp +
+                    WITHDRAWAL_DELAY,
+            "Cooldown period not over"
```
sunnyvempati, line 90: custom error would be good here to keep things consistent

packages/protocol/src/bridge/abstract/BondManagement.sol
```diff
@@ -163,18 +169,21 @@ abstract contract BondManagement is ReentrancyGuard, IBondManagement {
         address bondToken
     ) external virtual nonReentrant returns (MessagingReceipt memory) {
         require(bondToken != address(0), "Bond token cannot be zero address");
+        BridgeTypes.WithdrawalIntent storage intent = withdrawalIntents[
+            msg.sender
+        ][bondToken];
+        require(
+            intent.amount > 0,
+            "No withdrawal intent"
+        );
```
sunnyvempati, line 178: same as above

PR #686, opened 2025-09-01
Prediction Market V2 (old good parlay revisited)

llamalm:


packages/protocol/src/parlay/interfaces/IParlayPool.sol
```diff
@@ -1,69 +1,113 @@
...
-     */
-    function withdrawParlayCollateral(uint256 tokenId) external;
+    function mint(
+        IParlayStructs.PredictedOutcome[] calldata predictedOutcomes,
+        address resolver,
+        uint256 makerCollateral,
+        uint256 takerCollateral,
+        bytes calldata makerSignature,
+        bytes calldata takerSignature,
+        uint256 mintExpirationTime,
```
noahlitvin, line 95: i think we can remove this?

PR #1557, opened 2026-04-15
refactor volume semantics around similarMarketVolume

noahlitvin:
## Summary
- remove the legacy `volume` sort path and make `similarMarketVolume` the single volume sort, with windowed and filtered behavior controlled by `similarMarketVolumeWindow` and `excludeExtremeOdds`
- add explicit `minSimilarMarketVolume`/`maxSimilarMarketVolume` filters that apply independently of sort field and align resolver SQL + Prisma fallback behavior
- rename `excludeLowOdds` to `excludeExtremeOdds` across API schema, SDK query contracts, app hooks/components, and resolver tests

## Test plan
- [x] `pnpm --filter @sapience/api run generate-types`
- [x] `pnpm --filter @sapience/api run test -- QuestionsResolver.test.ts`
- [x] `pnpm --filter @sapience/api run test`
- [x] `pnpm --filter @sapience/sdk run build:lib`

Made with [Cursor](https://cursor.com)


nostradamus-agent (changes requested):
Found two blocking regressions:

1. `minSimilarMarketVolume` / `maxSimilarMarketVolume` aren't included in `hasConditionFilters` inside `QuestionsResolver`, so the SQL phase still uses unfiltered group totals when those are the only active filters. That misorders grouped results and can make pagination sparse because zero-match groups get counted by `LIMIT/OFFSET` and only disappear later in the reconstruction step.

2. `PUT /admin/conditions/volume` now only accepts `similarMarketVolume*` keys, but `packages/market-keeper/src/prices-and-1d-7d-volume/index.ts` still submits `volume24h` / `volume7d`. Those updates become silent no-ops, so the bootstrap 24h/7d volume refresh stops working until `refresh-volume` overwrites it later (if it runs at all).

7ayd (approved):

PR #1561, opened 2026-04-16
Staging to main 

7ayd:


nostradamus-agent (changes requested):
Found two app-side blockers in this staging -> main rollup:

1. `packages/app/src/components/markets/MarketsPage.tsx` sends `similarMarketVolumeWindow` as `1hFiltered/4hFiltered/...` when "Filter Volume" is enabled, but `packages/app/src/components/markets/QuestionsTable.tsx` still renders and client-sorts the unfiltered window values via `volumeWindowRef.current ?? '24h'`. That means the backend is sorting/filtering on filtered volume while the column still shows raw volume, so the UI lies about what it's ordering by.

2. This removes `volume` from the GraphQL `QuestionSortField` enum, but `packages/app/src/components/markets/MarketsPage.tsx` still restores `sapience.markets.sortField` straight from sessionStorage through `useSessionState` with no migration/validation. Existing users who previously picked Volume sorting will keep `"volume"` in storage and send an invalid enum on the next query after deploy.

I’m not approving this one until those are fixed.

nostradamus-agent:
Opened #1562 against `staging` with the app-side fixes for the two blockers I flagged here. Once that lands, this promotion PR should pick up the patch automatically on the next sync.

nostradamus-agent (approved):
Reviewed the staged changeset and the relevant checks are green on the current staging head.

PR #1567, opened 2026-04-18
feat(market-keeper): always include Maine and Michigan

ukitta555:
## Summary
Add Maine and Michigan to the `ALWAYS_INCLUDE_PATTERNS` in market-keeper so markets mentioning these states are always included regardless of volume thresholds.

## Test plan
- [ ] Verify markets with "Maine" or "Michigan" in the question bypass volume filters


nostradamus-agent (approved):
Reviewed the single-file market-keeper change. The new ALWAYS_INCLUDE_PATTERNS entries are scoped correctly, and the market-keeper check is green on the current head.

PR #1569, opened 2026-04-18
[1/4] Add GraphQL contract test suite

noahlitvin:
## Summary

First in a **four-PR stack** that migrates the API from `type-graphql` + `typegraphql-prisma` to an SDL-first GraphQL stack, then bumps Prisma 6.10 → 7.7. This PR is pure test infrastructure — it adds the contract suite **while the old stack is still running**, so the remaining PRs can use it as a regression gate.

## Stack

- **→ [1/4] Add GraphQL contract test suite** (this PR)
- [2/4] Migrate GraphQL stack to SDL-first
- [3/4] Drop contract-test subprocess harness + register custom scalars
- [4/4] Bump Prisma 6.10.1 → 6.19.3 → 7.7.0

Each PR targets the previous one's branch. When this lands, GitHub will auto-retarget [2/4] against `staging`.

## What's in here

Four commits:

1. `test(api): add GraphQL contract test suite` — tooling under `packages/api/test/contract/` that captures SDL, introspection, 12 frontend operation responses, and 27 orphan root-field error shapes as snapshots.
2. `test(api): capture GraphQL contract baseline snapshots` — first capture against the current Railway source DB. 17k-line introspection.json + per-operation snapshots under `test/contract/__snapshots__/`.
3. `test(api): exclude contract snapshots from prettier` — the snapshots serialize via `JSON.stringify(x, null, 2)` but lint-staged was reformatting them into compressed-array form on commit, causing every subsequent test run to false-fail. Fixed by adding the snapshots dir to `.prettierignore`.
4. `test(api): exclude contract suite from default vitest run` — the contract suite needs `TEST_GRAPHQL_URL` from its own `globalSetup`; running it under the default vitest config without that setup hard-errors. Added `test/contract/**` to `vitest.config.ts` exclude list.

## Why this PR first

The stack's whole safety argument is "swap the schema-builder without changing observable behaviour." Without these tests landed on `staging`, [2/4] has nothing to prove against. Land this, then the follow-ups can individually point at the contract suite passing 42/42 as evidence of non-regression.

## Test plan

- [ ] CI runs the default unit suite (contract is excluded, as it should be)
- [ ] `pnpm --filter @sapience/api run test:contract` passes 42/42 locally against a DB matching the captured fixture


llamalm:
One thing I noticed is that a bunch of 'snapshots' that are used in tests to see the actual result are empty, so the test are actually testing for no-returns

PR #1582, opened 2026-04-20
Staging -> Main

ukitta555:


packages/api/src/workers/indexers/__tests__/positionTokenTransferIndexer.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    removed: false,
+    topics: [] as unknown as [],
+    transactionHash:
+      (overrides.transactionHash as `0x${string}`) ??
+      (('0x' + 'ab'.repeat(32)) as `0x${string}`),
+    transactionIndex: 0,
+  };
+}
+
+const TOKEN_INFO_MAP = new Map<
```
nostradamus-agent, line 63: This is invalid TS syntax: `const TOKEN_INFO_MAP=*** Map<...>`. Looks like a paste/formatting artifact. It needs to be `new Map<...>(...)` or similar before this can merge.

nostradamus-agent (changes requested):
Found a blocker in the new test file, so I can't approve this one yet. `packages/api/src/workers/indexers/__tests__/positionTokenTransferIndexer.test.ts` has invalid TypeScript (`const TOKEN_INFO_MAP=*** Map<...>`), which won't parse.

nostradamus-agent (approved):
Re-checked the raw blob for the test file and my earlier blocker was wrong — the line is `const TOKEN_INFO_MAP = new Map<...>`, not invalid syntax. The idempotency/indexer changes look good and the relevant checks are green.

PR #1583, opened 2026-04-20
fix: leaderboard PnL math and vault balance display

noahlitvin:
## Summary

- Rewrite leaderboard PnL aggregation to use deterministic per-prediction PnL driven by `Picks.result` + actual collateral from `Prediction`, replacing the broken `Claim.collateralPaid − Claim.tokensBurned` formula that evaluated to 0 for most winning redemptions in the parimutuel model.
- Show true vault AUM on `/vaults` by summing `getTotalLiquidValue()` + deployed collateral, since the contract call intentionally excludes position tokens per its docstring.
- Rename the leaderboard column from "Realized Profit" to "Profit" — the metric always included unrealized-from-resolved positions, so the old label was misleading.
- Add `backfillVaultFlows.ts`, a one-shot chain reader for `PendingRequestProcessed` + `EmergencyWithdrawal` events on the current vault. `vault_flow_event` has no active indexer, so any consumer (airdrop-gain calc, flow charts) was reading stale pre-redeployment data.

## Why the leaderboard was wrong

`Claim.tokensBurned` in the parimutuel escrow equals the *full pot share*, not the user's individual stake — per `packages/protocol/CLAUDE.md`, both sides receive position tokens equal to `predictorCollateral + counterpartyCollateral`. So for every winning full redemption the contract correctly emits `collateralPaid == tokensBurned`, and the previous formula collapsed to PnL = 0. The fix keys off `Picks.result` and the actual collateral fields on `Prediction`, which also makes the sum of user PnL reconcile with `-vault_cumulative_pnl` (verified: sum = 0.0000 USDe exactly in prod).

## Why the vault balance was wrong

`getTotalLiquidValue()` returns `(liquidAssets, unconfirmedDeposits)` and intentionally excludes the value of position tokens held by the vault. The UI was rendering only `liquidAssets` while labelling it "VAULT BALANCE" and treating the "deployed" figure as a subset of it. Summing the two yields true AUM.

## Verified against prod (local API pointed at prod DB)

Top of the leaderboard before vs after:

| Address | Before | After |
|---|---|---|
| `0x207745a1…` | −6.09 USDe (rank 35) | **+1,274.91 USDe (rank 1)** |
| `0xFa80D2…` | +1.65 USDe (rank 10) | **+752.45 USDe (rank 2)** |
| Current vault `0x1f5fF6…` | — | +311.04 USDe (matches chart within ~28 USDe) |
| Sum of all user PnL | — | **0.0000 USDe** (perfect zero-sum) |

Dry-run of the backfill script found 1 legitimate deposit event on the current vault (7,923 USDe on 2026-03-26, right after redeployment); combined with ~335 USDe of realized PnL this reconciles with the fixed UI's ~8,258 USDe total AUM.

## Followups (not in this PR)

- Run `backfillVaultFlows.ts` against prod with write credentials (safe: idempotent on `(chainId, tx, logIndex)`).
- Stand up a live vault-flow indexer so `vault_flow_event` stays current.
- Operational recovery: ~586 USDe of winning position tokens are held by legacy vault `0x0f246f…` and never redeemed — the external redemption service appears scoped to the current vault only.

## Test plan

- [ ] Hard-reload `/leaderboard`; confirm `0x207745…` is rank 1 at ~1,275 USDe and the column header reads "Profit" (not "Realized Profit").
- [ ] Hard-reload `/vaults`; confirm "VAULT BALANCE" now reads ~8,258 USDe (liquid 6,122 + deployed 2,136) and the deployed percentage matches.
- [ ] Run `pnpm --filter @sapience/api exec tsx scripts/backfillVaultFlows.ts --dry-run` and verify the printed events look sensible.
- [ ] Smoke-test `accountProfitRank` GraphQL query for `0x207745…` and `0xFa80D2…` return the new numbers.

BY @ukitta555 -> TODO: fix burn SQL exclusion (too aggressive right now); think what to do with burns in general when it comes to PnL

### DIFF
--- a/packages/api/scripts/backfillVaultFlows.ts
+++ b/packages/api/scripts/backfillVaultFlows.ts
@@ -1,0 +1,190 @@
+/**
+ * One-shot backfill for vault_flow_event on the current vault deployment.
+ * Reads PendingRequestProcessed + EmergencyWithdrawal events from chain.
+ *
+ * Why: no indexer writes to this table today. The rows that exist predate the
+ * current vault redeployment (2026-03-25, block 3890223). Running this lets
+ * analytics queries (airdrop gains, cumulative flows) reflect reality.
+ *
+ * Usage:
+ *   pnpm --filter @sapience/api exec tsx scripts/backfillVaultFlows.ts
+ *   pnpm --filter @sapience/api exec tsx scripts/backfillVaultFlows.ts --chainId 5064014
+ *   pnpm --filter @sapience/api exec tsx scripts/backfillVaultFlows.ts --dry-run
+ *
+ * --dry-run prints what would be upserted without touching the database. Use it to
+ * verify event discovery and data shape before running with write credentials.
+ */
+import { parseAbiItem } from 'viem';
+import prisma from '../src/db';
+import { getProviderForChain } from '../src/utils/utils';
+import { contracts } from '@sapience/sdk/contracts';
+
+const CHAIN_ID = (() => {
+  const eq = process.argv.find((a) => a.startsWith('--chainId='));
+  if (eq) return Number(eq.split('=')[1]);
+  const idx = process.argv.indexOf('--chainId');
+  if (idx > 0 && process.argv[idx + 1]) return Number(process.argv[idx + 1]);
+  return 5064014;
+})();
+
+const DRY_RUN = process.argv.includes('--dry-run');
+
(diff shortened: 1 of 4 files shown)


ukitta555 (approved):
TODO: fix burn SQL exclusion (too aggressive right now); think what to do with burns in general when it comes to PnL

PR #1587, opened 2026-04-21
fix/update conditions url 

7ayd:


nostradamus-agent (approved):
Reviewed the market-keeper changes. The PR fixes the broken Polymarket condition-group URLs for new writes and adds a backfill path for existing stale group metadata. test-market-keeper is green on the current head.

PR #1589, opened 2026-04-21
add deploy-pyth-vault + generic SDK address sync, register singleLegVault

llamalm:
## Summary

Adds a `deploy-pyth-vault` command to `deploy.sh` that deploys a dedicated `PredictionMarketVault` instance for the Pyth binary-options bot and syncs its address into the SDK's `addresses.ts`. Also ships a generic SDK address sync helper reusable for any `ChainAddressMap` export, and registers the already-deployed `singleLegVault` mainnet address.

## Changes

- **`packages/sdk/contracts/addresses.ts`**
  - New export `pythPredictionMarketVault: ChainAddressMap` (empty bootstrap — populated by `deploy-pyth-vault`).
  - New export `singleLegVault: ChainAddressMap` populated with the existing Ethereal mainnet deployment: `0x1b03b3f20caa6fc8cc7f6d9ae73d634804fe7f59` (block `3915329`, deployed 2026-03-26). Sourced from the on-chain creation tx via Blockscout.
  - Both exports added to the `contracts` object.

- **`packages/protocol/src/scripts/deploy/update-sdk-vault.mjs` (new)** — generic helper invoked by `deploy.sh` after any vault-style deploy:
  - `--exportName <name>` targets any `ChainAddressMap` export (e.g. `pythPredictionMarketVault`, `singleLegVault`, future vaults).
  - `--chainId`, `--address` required; `--blockCreated`, `--deployDate` optional (deployDate defaults to today).
  - On redeploy, the previous `{ address, blockCreated }` is prepended to the chain's `legacy` array (most recent first); pre-existing legacy entries are preserved. Same-address replacement is a no-op for legacy (handles CREATE2 reuse).
  - Errors out clearly if the export isn't bootstrapped in `addresses.ts`.

- **`packages/protocol/src/scripts/deploy/deploy.sh`** — new `deploy_pyth_vault()` function + `deploy-pyth-vault` command + help text:
  - Reuses the existing `DeployVault.s.sol` forge script but overrides `VAULT_*` env (`VAULT_MANAGER`, `VAULT_NAME`, `VAULT_SYMBOL`, `VAULT_EXPIRATION_TIME`) from `PYTH_VAULT_*` equivalents just for that invocation.
  - **Save/restore** of the original `VAULT_*` values around the run so subsequent `deploy-vault` calls in the same shell aren't affected (inline assignments leak into parent shell when the target is a bash function).
  - Snapshots `cast block-number` before deploying to record `blockCreated` in the SDK.
  - Writes `PYTH_VAULT_ADDRESS` to `config.json`, `PythPredictionMarketVault` to `deployments.json`, and calls `update-sdk-vault.mjs --exportName pythPredictionMarketVault` to sync `addresses.ts`.
  - Prints next-step hints (seed vault, `approveFundsUsage`, commit SDK diff).

## Test plan

Verified the updater end-to-end with 4 scenarios against a local checkout:

- [x] Fresh insert into empty `singleLegVault` — `Added singleLegVault[5064014] = ...`
- [x] Redeploy — old address moves to `legacy`, new becomes primary. Log prefix: `Replaced (old address moved to legacy)`.
- [x] Third deploy — both previous addresses kept in `legacy`, most recent first.
- [x] Add second chain (testnet) — mainnet entry + its `legacy` array untouched.
- [x] Missing args, unknown `--exportName`, invalid `--deployDate` all error out clearly.
- [x] `bash -n deploy.sh` clean. `save/restore` of `VAULT_*` validated in a sandbox: original values (including unset) are restored exactly after the function call.

Manual deploy run against a real testnet/mainnet forge broadcast is pending — would appreciate a review of the deploy.sh changes before running that.


nostradamus-agent (approved):
Reviewed the 4-file staging PR. The new deploy-pyth-vault flow is scoped cleanly, the SDK address sync helper is sane, and the SDK additions type-check. I also sanity-checked deploy.sh with bash -n. Vercel checks are green and I did not find anything blocking.

PR #1591, opened 2026-04-21
fix: redeploy testnet PythConditionResolver with correct pythLazer

llamalm:
## Summary

Redeploys `PythConditionResolver` on Ethereal testnet because the previous deployment (`0xF56d26C58496A07DF4882F52c56E3cadBFB87729`) was wired to an incorrect `pythLazer` verifier (`0xACeA761c27A909d4D3895128EBe6370FDE2dF481`), which caused the market-keeper's `verifyUpdate()` call to revert with `invalid signer` on every Pyth condition it tried to settle.

The new resolver (`0x7BD534c96b78A5d6C10701714F29EA4f1587B64D`) points at the correct testnet `pythLazer` per `testnet/config.json` (`0x4D4772F06c595F69FB57039599a180536FDE8245`). Since `pythLazer` is declared `immutable` in `PythConditionResolver.sol:60`, a redeploy is the only remediation.

## Changes

- `packages/protocol/src/scripts/testnet/config.json` — `PYTH_CONDITION_RESOLVER_ADDRESS` updated to the new address.
- `packages/sdk/contracts/addresses.ts` — `pythConditionResolver[13374202]` updated; the previous `0xF56d…` address moved into `legacy` (preserving the pre-existing legacy entries). Generated via `update-sdk-vault.mjs --exportName pythConditionResolver --chainId 13374202 ...`.

## Verification

```bash
# New resolver points at the correct pythLazer
$ cast call 0x7BD534c96b78A5d6C10701714F29EA4f1587B64D "pythLazer()(address)" --rpc-url https://rpc.etherealtest.net
0x4D4772F06c595F69FB57039599a180536FDE8245   # matches testnet/config.json:26
```

## Operational follow-ups (not in this PR)

- [ ] Update `EXPECTED_RESOLVER_ADDRESS` in the `pyth-auction-bidder` Railway staging service to the new resolver, then redeploy that service.
- [ ] Any positions already minted against the old resolver (`0xF56d…`) on testnet will remain unsettleable via the keeper — irrelevant for testnet but worth noting.
- [ ] Extend `deploy-pyth-resolver` in `deploy.sh` to call `update-sdk-vault.mjs` automatically after a successful deploy, analogous to how `deploy-pyth-vault` already does. Avoids the same SDK-drift bug from recurring. Can be a separate follow-up.


nostradamus-agent (approved):
Reviewed the two-file redeploy follow-up. No blockers from me.

Verified:
- `packages/protocol/src/scripts/testnet/config.json` points `PYTH_CONDITION_RESOLVER_ADDRESS` at the new testnet resolver `0x7BD534c96b78A5d6C10701714F29EA4f1587B64D`
- `packages/sdk/contracts/addresses.ts` updates `pythConditionResolver[13374202]` and keeps the previous resolver in `legacy`
- On-chain, the new resolver returns the expected testnet `pythLazer()` = `0x4D4772F06c595F69FB57039599a180536FDE8245`
- On-chain, the old resolver still returns the incorrect verifier `0xACeA761c27A909d4D3895128EBe6370FDE2dF481`, matching the root-cause analysis

Local signal:
- `pnpm --filter @sapience/sdk run lint`

Approving — this fixes the resolver address drift cleanly without breaking historical lookups for the old testnet resolver.

PR #1595, opened 2026-04-21
Staging → main

nostradamus-agent:
## Summary

Promotes the current `staging` branch to `main`.

- `staging` is ahead of `main` by 20 non-merge commits.
- `staging` is currently behind `main` by 1 commit, so GitHub may report conflicts until that delta is reconciled.

## Included commits

- fix(tags): normalize first-letter case on ingest + backfill
- fix(keeper): settle conditions on legacy resolvers (#1565)
- fix(analytics): correct OI, TVL, volume metrics and live candle (#1568)
- chore(api): restore schema.graphql + graphql.ts to match CI regeneration
- fix(app): assorted markets & vaults polish (#1575)
- fix(app): base Protocol TVL on escrow balance, not open interest (#1576)
- fix(app): render aggregate Position (not a specific Prediction) from positions list (#1577)
- fix(api): make indexer handlers idempotent to stop OI/balance drift (#1580)
- fix(api): allow past endTimes in batch-create conditions (#1581)
- feat(app,api): sortable positions table columns (#1579)
- feat(app): tighten bid-estimate hints and Pyth prediction line UI (#1584)
- feat(market-keeper): one-off script to add Polymarket events by slug (#1586)
- fix-shortnames
- add deploy-pyth-vault + generic update-sdk-vault helper, register singleLegVault
- testnet deployment
- redeploy PythConditionResolver on testnet with correct pythLazer address
- fix(ui): raise Pyth prediction form inline layout to 1500px
- fix: leaderboard PnL math and vault balance display (#1583)
- fix(activity): scope Related Activity by pickConfigId and conditionId (#1592)
- chore(api): remove dead code and consolidate duplicated tests (#1594)


packages/market-keeper/scripts/add-events.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  if (!privateKey) {
+    console.error('ADMIN_PRIVATE_KEY is required for submission (use --dry-run to preview)');
+    process.exit(1);
+  }
+
+  await submitToAPI(apiUrl, privateKey, sapienceData);
+
+  if (sapienceData.metadataUpdates.length > 0) {
```
nostradamus-agent, line 134: `groupMarkets()` returns both `metadataUpdates` and `groupMetadataUpdates`, but this script only submits the former. The normal generate path also calls `submitGroupMetadataUpdates(...)`, so manual event backfills can leave existing group title/description/tag metadata stale. This needs the same follow-up submit here.

nostradamus-agent (commented):
Review pass done. I found one blocker before this should merge.

Blocker:
- `packages/market-keeper/scripts/add-events.ts` runs the same grouping pipeline as `generate`, but it only submits `metadataUpdates` and never submits `groupMetadataUpdates`. That means manual event backfills can leave existing condition-group metadata stale even though the normal generate path updates it.

Non-blocking note:
- This new write script also skips the usual production confirmation guard for `api.sapience.xyz`. I'd add that before someone fat-fingers prod from a non-prod shell.

I couldn't submit a formal `REQUEST_CHANGES` because GitHub doesn't allow that on my own PR, so I'm leaving this as a review comment instead.

PR #1603, opened 2026-04-22
chore(api): remove unused NOAA_TOKEN from .env.sample

llamalm:
## Summary
- Remove `NOAA_TOKEN=` from `packages/api/.env.sample`. It is not referenced anywhere in the codebase.
- Dropped as part of the security inventory pass (sapience-docs PR #4): shipping env-var templates that nothing reads makes the sensitive-data surface look larger than it actually is and misleads people onboarding new services.

## Verification
- `grep -ri "NOAA" --exclude-dir=node_modules --exclude-dir=.git --exclude-dir=dist --exclude-dir=out --exclude-dir=.next` returns zero matches across the whole monorepo after this change.

## Test plan
- [ ] Confirm no service (`packages/api`, `packages/market-keeper`, etc.) reads `process.env.NOAA_TOKEN` at runtime
- [ ] Confirm Railway / Vercel env configs for the sapience project do not rely on the variable


nostradamus-agent (approved):
Verified current PR head and check status. Relevant checks for this API-only cleanup are green.

PR #1632, opened 2026-04-27
feat(app): inline Bungee bridge for funding the Sapience smart account

noahlitvin:
## Summary

Replaces the external **Get USDe** / **Bridge to Ethereal** links in the Fund Account flow with an in-app **Deposit** dialog that bridges USDe / ETH / USDC / USDT from Arbitrum or Mainnet straight to the smart account on Ethereal via Bungee's deposit-address API.

- Destination is locked to `smartAccountAddress`; refund target is the user's EOA on the source chain
- Bungee deposit-address flow needs **no token approval** — user signs one tx (ERC-20 `transfer` or native send) on the source chain
- ~40-200s ETA depending on route; quote auto-refreshes every 30s; status polled every 5s until terminal
- The `Deposit Ethereal USDe to get started` banner now opens this dialog instead of linking out to Stargate

## What's in this PR

**New**
- `lib/bungee.ts` — quote/status helpers, supported source chains + tokens, status-code utilities
- `BungeeBridgePanel.tsx` — single \`From\` picker (chain × token) with balances, USD values, auto-select highest-balance combo, per-token-decimal amount handling
- `FundDialog.tsx` + `FundDialogContext.tsx` — context-driven dialog so both the balance pill button and the LowBalanceBanner can open it (mirrors `ConnectDialogContext`)

**Wired**
- `LowBalanceBanner.tsx` swapped from external Stargate link to `openFundDialog()`
- `CollateralBalanceButton.tsx` rewired to `useFundDialog()` and shed ~150 lines of inline bridge UI (the Withdraw dialog is unchanged)
- `providers.tsx` adds `mainnet` to wagmi config so the bridge can switch chains and read source-chain balances

**Drive-by**
- `MarketsPage.tsx`: tweak `ExampleCombos` top spacing — `mt-4 lg:mt-0` → `mt-4 xl:mt-0` so it doesn't collapse to zero before the two-column layout actually appears

## Verified

- Address checks: every (token, chain) entry verified against Bungee's quote API. Note: Arbitrum USDT (0xFd086b…) returns "USDT0" from Bungee — that's their relabel of the canonical Tether-on-Arbitrum contract, which is what users have in their wallets. We display "USDT" in the chip.
- API endpoint: \`https://public-backend.bungee.exchange\` (sandbox — no key, fine for dev/staging). For prod mainnet ramp, switch to \`backend.bungee.exchange\` with domain whitelisting or proxy through our backend with a key.
- Type-check + lint clean

## Test plan

- [ ] Open the Deposit dialog from the **Fund Account** button (smart-account mode, zero balance)
- [ ] Open it from the **Deposit Ethereal USDe to get started** banner (low-balance state)
- [ ] Pick a non-default combo (e.g. ETH on Mainnet); verify the amount auto-fills to that EOA's balance, capped at 6 decimals
- [ ] Click 50% / Max — value updates
- [ ] Sign a real bridge from Arbitrum USDe → Sapience smart account; confirm chain-switch UX, source-chain tx, and status progression to Delivered
- [ ] Confirm balance pill updates after delivery (auto-refetch fires)
- [ ] Verify zero-balance combos are hidden (with selected combo as fallback)


packages/app/src/lib/bungee.ts
```diff
@@ -0,0 +1,311 @@
...
+      'https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/info/logo.png',
+  },
+  {
+    chainId: 42161,
+    name: 'Arbitrum',
+    iconUrl:
+      'https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/arbitrum/info/logo.png',
+  },
+  {
+    chainId: 8453,
```
nostradamus-agent: ⚠️ This exposes Base/BSC/HyperEVM in the source picker, but `providers.tsx` only registers Ethereum + Arbitrum + Ethereal chains/transports. `useReadContracts`, `useBalance`, and especially `switchChainAsync({ chainId: 8453/56/999 })` won't have wagmi config for these routes. Either add the missing chain definitions/transports to the wagmi config, or keep this allowlist to only chains the app can actually read/switch/send on.

packages/app/src/components/layout/BungeeBridgePanel.tsx
```diff
@@ -0,0 +1,758 @@
...
+  useEffect(() => {
+    if (!sourceToken) return;
+    const key = `${selectedKey}|${eoaAddress ?? ''}`;
+    if (lastAutoFillKey.current === key) return;
+    const balanceLoaded = sourceToken.isNative
+      ? !nativeBalancesLoading
+      : erc20BalancesData !== undefined;
+    if (!balanceLoaded) return;
+    setAmount(
+      sourceBalance > 0n
```
nostradamus-agent: ⚠️ For native source tokens this autofills the entire ETH/BNB/HYPE balance, and the Max button does the same below. `handleBridge()` then sends Bungee's `txData.value`, so the wallet has no native asset left to pay gas and the deposit fails as insufficient funds. Native-token paths need to reserve gas / cap max below the full balance, or avoid auto-filling Max for native assets.

nostradamus-agent (changes requested):
Found two blockers in the bridge flow. The shape is good, but right now some advertised source routes can't work in wagmi, and native-token max deposits will try to spend the gas token itself. 🔮

nostradamus-agent (approved):
Re-reviewed after the fix commit. The two blockers are addressed:

- picker is now scoped to source chains the app actually configures in wagmi
- native-token Max/autofill reserves source-chain gas instead of spending the full gas-token balance

Local checks are green: app lint, app type-check, app vitest suite (528 tests).

nostradamus-agent (commented):
## Code Review Summary

**Verdict: Comment only** — no hard blockers found, but one concurrency sharp edge is worth tightening.

### ⚠️ Warning
- `packages/api/src/workers/indexers/predictionMarketEscrowIndexer.ts:528` — `Event` is now written after the handler, but `PredictionCreated` is not fully idempotent under a duplicate concurrent pass. Two workers can both miss the `Event` row and both enter `processPredictionCreated`; one wins `tx.prediction.create`, the other hits the `Prediction.predictionId` unique constraint and gets captured by the outer error handler/Sentry instead of short-circuiting cleanly. The transaction rollback prevents double-applying the OI writes, so this is not data corruption, but it turns a normal background-worker/reconciler race into noisy failed processing. I’d catch/ignore the prediction `P2002` (or otherwise make `PredictionCreated` structurally idempotent) before relying on the “handlers are idempotent” invariant here.

### ✅ Verified
- `pnpm --filter @sapience/app exec vitest run ...` for new bungee/withdraw/position/format tests — 102 passed
- `pnpm --filter @sapience/app run lint`
- `pnpm --filter @sapience/sdk run test` — 649 passed
- `pnpm --filter @sapience/api run lint`
- Targeted API protocolStats/analytics/query-cache tests passed; tests needing generated Prisma could not run locally because this sandbox has Node 22.11.0 and Prisma 7 requires the repo-pinned 22.12.0.

PR #1636, opened 2026-04-28
Staging → main

nostradamus-agent:
## Summary
Promote `staging` to `main` with user-facing analytics/profile improvements, protocol-stats correctness/backfill changes, SDK/query updates, and a couple of ops/tooling updates.

## App
- show realized PnL for closed positions in profile views, including synthetic secondary-market sold rows
- add analytics charts for open interest by category and time to resolution
- improve protocol analytics chart handling for sub-daily snapshot cadences
- anchor the vault PnL chart Y-axis at zero when all values are non-negative
- improve Sentry capture for RPC, paymaster, and session-key failures

## API
- rework protocol-stats snapshots/backfills to be block-pinned and interval-aware
- add GraphQL analytics resolvers for `openInterestByCategory` and `openInterestByTimeToResolution`
- add GraphQL operation timing logging
- extend position data with realized PnL and align position-count behavior with filtered rows

## SDK
- switch analytics consumers from `dailyPnL` / `dailyVolume` to `periodPnL` / `periodVolume`
- expose new analytics query helpers
- export shared auction helpers moved out of the app
- regenerate GraphQL types to match the new API surface

## Infra / ops
- pin Node to `22.12.0` for Prisma/tooling stability
- move Polymarket CLOB integrations to `clob-v2.polymarket.com`

## Rollout notes
- most user-visible changes are in profile realized-PnL views, new analytics charts, and cleaner vault/protocol chart behavior
- biggest rollout risk is the broad protocol-stats/backfill change set, so that area deserves extra scrutiny in review and post-deploy validation


packages/api/src/helpers/protocolStats.ts
```diff
@@ -465,57 +679,123 @@ async function upsertProtocolStatsSnapshot(
...
-    `[ProtocolStats] Escrow: ${formatUnits(escrowBalance, 18)} USDe (all contracts)`
+    `[ProtocolStats] Escrow: ${formatUnits(escrowBalance, 18)} USDe (V2 primary + past deploys)`
   );
 
-  // Calculate vault PnL
-  const pnlResult = await calculateVaultPnL(chainId);
+  // DB-derived aggregates — pass the same snapshot timestamp so these are also
+  // evaluated at "state as of timestamp", matching the on-chain reads above.
+  const [vaultDeployed, pnlResult, flowsResult] = await Promise.all([
+    fetchVaultDeployedAtBlock(chainId, blockNumber, timestamp),
```
nostradamus-agent, line 790: This is the blocking bug for the rollout: the historical on-chain reads are block-pinned to the vault active at that block, but the DB-derived `vaultDeployed` / PnL aggregation still keys off the current primary vault address only. That can corrupt pre-redeploy protocol-stats snapshots.

nostradamus-agent (commented):
Blocking comment because I opened this PR and GitHub won't let me formally request changes on my own rollout PR.

Main blocker:
- `packages/api/src/helpers/protocolStats.ts` now picks the historically correct vault contract for on-chain reads, but the DB-derived helpers still aggregate against the current primary vault address only. That means pre-redeploy snapshots can mix legacy vault balances with current-vault `vaultDeployed` / PnL / derived airdrop data.

Concretely:
- `fetchVaultDeployed()` hardcodes `contracts.predictionMarketVault[chainId]?.address` at the top of the helper.
- `calculateVaultPnL()` does the same.
- `buildVaultAggregator()` also seeds from the current primary only.
- Later, `computeAndStoreProtocolStats()` and `backfillProtocolStats()` use those DB aggregates alongside block-pinned historical balance reads.

So snapshots before the 2026-03-25 vault redeploy can be internally inconsistent. I would not ship this to `main` until the DB-side historical aggregation uses the address active at the snapshot block/timestamp (or unions current + legacy addresses in a way that preserves historical correctness).

PR #1652, opened 2026-04-29
feat(app): show Pyth-resolved markets as Prices in OI by Category

noahlitvin:
## Summary
- Collapse the per-asset-class \`prices-*\` categories (\`prices-crypto\`, \`prices-commodities\`, \`prices-equity\`) into a single \"Prices\" slice on the Open Interest by Category chart, rendered with the Pyth mark.
- Extract the merge as \`collapsePriceCategories\` in \`market-helpers\` with unit tests so other surfaces can reuse the same rule.
- Add a \`prices\` FOCUS_AREA backed by a new \`--category-8\` token so the slice/badge color is deterministic instead of hash-derived.
- Key chart cells/list items by slug.

## Test plan
- [ ] On staging, open the analytics page and confirm the Open Interest by Category chart shows a \"Prices\" slice with the Pyth icon when the protocol has any Pyth-resolved (\`prices-*\`) open interest.
- [ ] Confirm non-price categories continue to render with their existing icons/colors.
- [ ] \`pnpm --filter @sapience/sdk run build:lib && pnpm --filter @sapience/app run test\`.
- [ ] \`pnpm --filter @sapience/app run type-check\`.


packages/app/src/lib/constants/focusAreas.ts
```diff
@@ -77,6 +78,13 @@ export const FOCUS_AREAS: FocusArea[] = [
     color: 'hsl(var(--category-6))',
     Icon: Tv,
   },
+  {
+    id: 'prices',
```
nostradamus-agent, line 82: This makes `prices` look like a real top-level category everywhere we iterate `FOCUS_AREAS`, but in this PR it is only a synthetic aggregate for analytics. For example, `/forecasts` renders its category tabs from `FOCUS_AREAS` (`ForecastPageContent.tsx`), while `Comments.tsx` still filters with `comment.category === selectedFilter`, so clicking the new `Prices` tab will not match the persisted `prices-crypto` / `prices-equity` / `prices-commodities` comments. I would keep `prices` local to the chart layer, or normalize `prices-*` ↔ `prices` everywhere that consumes these categories.

packages/app/src/components/analytics/OpenInterestByCategoryChart.tsx
```diff
@@ -61,8 +65,8 @@ export default function OpenInterestByCategoryChart() {
             No open interest yet.
           </div>
         ) : (
-          <div className="flex-1 grid grid-cols-1 sm:grid-cols-[1fr_minmax(0,220px)] gap-4 items-center">
-            <div className="h-full min-h-[240px] max-h-[300px] mx-auto w-full">
+          <div className="flex-1 grid grid-cols-1 sm:grid-cols-[1fr_minmax(0,220px)] gap-6 items-stretch">
+            <div className="min-h-[260px] mx-auto w-full">
```
nostradamus-agent, line 69: This wrapper used to provide a real height (`h-full min-h-[240px] max-h-[300px]`). After this change the child `ResponsiveContainer` is `height="100%"` inside a parent with only `min-h-[260px]`, which is not a definite height for Recharts. That can leave the pie at 0px tall depending on layout. I would restore an explicit height (`h-[260px]`, or `h-full` with a definite-height parent).

nostradamus-agent (changes requested):
Two blockers before I'd merge this:

1. Adding `prices` to the global `FOCUS_AREAS` list leaks a synthetic analytics-only bucket into other category-driven UI. The forecasts page renders its tabs from `FOCUS_AREAS`, but the comments filter still matches exact persisted category slugs, so the new **Prices** tab won't show the existing `prices-*` forecasts/comments.
2. The pie chart wrapper no longer gives `ResponsiveContainer` a definite height. `min-h-[260px]` isn't enough for `height="100%"`, so this can collapse the chart while the legend still renders.

I did get the new unit test passing locally in the PR worktree (`market-helpers.test.ts`).

PR #1653, opened 2026-04-29
fix(app): canonical picks in mint payload + surface UserOp reverts

noahlitvin:
## Summary

- **Root-cause fix**: Multi-pick parlays were silently failing on-chain with `InvalidCounterpartySignature()` (selector `0x98595156`) for ~50% of orderings. `useAuctionStart` stored picks in the user's input order while the bidder/vault signed over the canonical (keccak256-sorted) order shipped by `prepareAuctionRFQ`. The on-chain `predictionHash` recomputed by `mint()` therefore diverged from what the counterparty signed. Fixed by writing `prepared.canonicalPicks` back into `lastAuctionRef.current.picks` so `buildMintRequestDataFromBid` produces a payload whose `predictionHash` matches the counterparty's signature.
- **Surface UserOp reverts**: `executeViaSessionKeyDefault` now awaits `waitForUserOperationReceipt` and throws on revert. Previously, a reverted UserOp returned silently and the UI hung on "INDEXING POSITION..." forever because the indexer never saw a position. Now the user gets a real error toast (`Transaction reverted onchain: <selector or reason>`).
- **Dialog UX**: The progress dialog opens immediately when the user clicks submit (in `SUBMITTING` stage) instead of after `submitPosition` resolves. On failure, dialog closes and progress resets so the user can retry.
- **Toast wording**: "Transaction reverted onchain" (no hyphen).
- **Regression test** in `transactionExecutor.test.ts` covers the new revert-surfacing behavior.

## Test plan

- [x] `pnpm --filter @sapience/app exec vitest run src/hooks/blockchain/transactionExecutor.test.ts` (38 passing)
- [x] Manually reproduced and confirmed fix: 2-pick parlay (Keiko Fujimori NO + Haley Stevens YES) submitted successfully against the mainnet `0x1f5fF6…` vault from a smart-account predictor. Same flow had reverted with `InvalidCounterpartySignature()` before this change.
- [ ] Verify on staging deploy that single-pick mints still work (canonicalization is a no-op there but worth confirming).
- [ ] Verify on staging that a forced revert still shows a clear toast and the dialog closes / lets the user retry.

### DIFF
--- a/packages/app/src/lib/auction/useAuctionStart.ts
+++ b/packages/app/src/lib/auction/useAuctionStart.ts
@@ -8,6 +8,7 @@ import {
   prepareAuctionRFQ,
   type SignableTypedData,
 } from '@sapience/sdk/auction/initiate';
+import { canonicalizePicks } from '@sapience/sdk/auction/escrowEncoding';
 import { useSettings } from '~/lib/context/SettingsContext';
 import { useSession } from '~/lib/context/SessionContext';
 import { toAuctionWsUrl } from '~/lib/ws/auctionUrl';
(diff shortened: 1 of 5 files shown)


packages/app/src/lib/auction/useAuctionStart.ts
```diff
@@ -495,9 +497,17 @@ export function useAuctionStart(options?: UseAuctionStartOptions) {
...
       }
 
-      // Store predictorDeadline on the auction ref so buildMintRequestDataFromBid can access it
+      // Store predictorDeadline on the auction ref so buildMintRequestDataFromBid can access it.
+      // Also overwrite picks with the canonical ordering used by prepareAuctionRFQ — bidders
+      // sign over canonical picks, so the mint payload must use the same order or
+      // verifyMintPartySignature reverts with InvalidCounterpartySignature.
       lastAuctionRef.current = {
         ...lastAuctionRef.current,
+        picks: canonicalPicks.map((p) => ({
```
nostradamus-agent: This fixes the mint payload path, but `currentAuctionParams` still holds the pre-canonicalized `params.picks` from earlier in `requestQuotes()`. `CreatePositionForm` reads `currentAuctionParams.picks` for `useValidatedBids` and for `OgShareDialogBase.expectedPicks`, so validation / index-tracking can still be out of sync with the canonical order the bidder signed. I think `currentAuctionParams` needs the same canonicalized picks update as `lastAuctionRef`.

nostradamus-agent (changes requested):
Found one blocker in the canonical-picks fix.

This updates `lastAuctionRef` to canonical order for mint submission, but the rest of the form still reads `currentAuctionParams.picks` for bid validation and share-dialog indexing. Since `currentAuctionParams` is still seeded from the original request order, those paths can keep using the wrong pick ordering even after this PR.

Targeted local signal: `src/hooks/blockchain/transactionExecutor.test.ts` passed. I did not treat app-wide type-check in this lightweight worktree as useful review signal because it throws unrelated workspace noise.

nostradamus-agent (approved):
Latest commits address my blocker. `requestQuotes()` now canonicalizes picks up front and seeds both `lastAuctionRef` and `currentAuctionParams` from the same canonical order, so mint submission, bid validation, and share-dialog/index tracking all agree on the prediction hash. I also reran the targeted transaction executor tests locally and they still pass.

PR #1654, opened 2026-04-29
feat(api): filter collateralTransfers by protocol exclusion

noahlitvin:
## Summary

- Adds `excludeProtocol: Boolean = false` arg to the `collateralTransfers` query, and removes the `@deprecated` directive (noah confirmed it can be modified, deprecation was incidental cleanup).
- When `excludeProtocol=true`, the resolver filters out any transfer whose `from` or `to` is a known protocol contract address (current + legacy across the full SDK `contracts` map for the requested chain). What remains is "external money flow" — third-party bridges (LayerZero OFT, Bungee), exchange withdrawals, peer-to-peer EOA transfers — i.e. what a "deposits & withdrawals" view wants to show.
- Default `false` preserves back-compat for any existing callers (admin/debug paths).

## Why this approach

The alternative was letting the FE pass explicit `to`/`from` allow-list arrays. Exclusion-by-protocol is cleaner because anything *not* in `addresses.ts` is by definition external — new bridge integrations don't need client updates. Stays out of bridge-protocol opinions (no LayerZero EID enrichment in this PR; that can be added later as a separate, bridge-agnostic concern when there's a concrete UX for it).

## Implementation notes

- New helper `getProtocolAddressesForChain(chainId)` in `@sapience/sdk/contracts` — flattens current + legacy addresses across every `contracts` export, deduped, lower-cased.
- Resolver adds `AND: [{ from: { notIn: ... } }, { to: { notIn: ... } }]` only when the flag is set and there are addresses to exclude on that chain (falls back to the existing query shape otherwise).

## Test plan

- [x] SDK helper unit tests: current address, legacy addresses, lower-casing, dedup, unknown chain returns `[]`
- [x] Resolver unit tests (TDD): default behavior unchanged, current escrow filtered, legacy escrow filtered, secondary-market escrow filtered, no-op on unknown chain, `excludeProtocol=false` ≡ omitted, address lower-cased, limit cap preserved
- [x] `pnpm --filter @sapience/sdk run test` (649/649)
- [x] `pnpm --filter @sapience/api run test` (444/444)
- [x] `pnpm --filter @sapience/api exec tsc --noEmit`
- [x] `pnpm --filter @sapience/api run lint`
- [x] `pnpm --filter @sapience/sdk run lint`

## Out of scope (deferred)

- LayerZero `dstEid` / destination-chain enrichment for outbound bridges. Will surface when there's a concrete UX requirement; should be done in a bridge-agnostic way (index outbound bridge events into their own table, expose `destinationChainId` separately).


packages/api/schema.graphql
```diff
@@ -1510,7 +1510,7 @@ type Query {
   closes(address: String, chainId: Int, pickConfigId: String, skip: Int! = 0, take: Int! = 50): [Close!]! @deprecated
   collateralBalance(address: String!, atBlock: Int, chainId: Int!): CollateralBalanceType!
   collateralBalanceHistory(address: String!, chainId: Int!, count: Int! = 12, intervalHours: Int! = 168): [CollateralBalanceSnapshotType!]!
-  collateralTransfers(address: String!, chainId: Int!, limit: Int! = 100, offset: Int! = 0): [CollateralTransferType!]! @deprecated
+  collateralTransfers(address: String!, chainId: Int!, excludeProtocol: Boolean = false, limit: Int! = 100, offset: Int! = 0): [CollateralTransferType!]!
```
nostradamus-agent, line 1513: Blocking: this changes the public GraphQL contract, but the contract snapshots weren't regenerated. `packages/api/test/contract/__snapshots__/schema.sdl.graphql` still has the old `collateralTransfers(address, chainId, limit, offset)` signature, and `introspection.json` will be stale for the same reason. Please refresh the contract snapshots along with the generated schema/types.

nostradamus-agent (changes requested):
Found one blocker before this is merge-ready.

The GraphQL contract changed (`collateralTransfers` now takes `excludeProtocol` and dropped the deprecation marker), but the committed contract snapshots were not refreshed. `packages/api/test/contract/schema.snapshot.test.ts` still snapshots `packages/api/test/contract/__snapshots__/schema.sdl.graphql`, and `introspection.test.ts` still snapshots `packages/api/test/contract/__snapshots__/introspection.json`. Neither file is updated in this PR, so the recorded public contract is stale.

Everything else in the implementation looks sane; this just needs the snapshot refresh before merge.

noahlitvin:
Pushed eb7b6b4b7 — surgical updates to both snapshot files for the new `excludeProtocol` arg only:

- `packages/api/test/contract/__snapshots__/schema.sdl.graphql` — adds the arg to the `collateralTransfers` line.
- `packages/api/test/contract/__snapshots__/introspection.json` — inserts the new arg in alphabetical position with `defaultValue: "false"`, type `Boolean` (nullable).

Kept the diff minimal so unrelated pre-existing snapshot drift (e.g. missing `vault*` fields, missing `@deprecated` annotations) isn't bundled into this PR's scope. Happy to do a separate cleanup PR for that if useful.

nostradamus-agent (approved):
Reviewed the PR diff against its recorded staging base.

Local signal I checked:
- `packages/sdk/contracts/__tests__/addresses.test.ts` passes in the PR worktree after building the SDK
- `packages/api/src/graphql/sdl/resolvers/queries/collateralBalance.test.ts` passes in the PR worktree
- GraphQL contract artifacts are updated too (`schema.graphql`, SDK graphql types, SDL/introspection snapshots), so this isn't drifting the public contract files

LGTM.

nostradamus-agent:
Re-reviewed the current PR head (`f94ba673`). No new code has landed since the approval — it's still the same 9-file diff (`excludeProtocol` resolver/helper/tests + GraphQL artifacts/snapshots).

I also re-checked the scary API/tooling failure locally: the `packages/api` Prisma/generate issue reproduces on clean `origin/staging`, so that looks like broader staging/toolchain drift rather than something introduced by this PR.

PR-specific signal still looks good to me:
- `packages/sdk/contracts/__tests__/addresses.test.ts` passes
- `packages/api/src/graphql/sdl/resolvers/queries/collateralBalance.test.ts` passes

So my verdict on this PR itself is still LGTM.

PR #1657, opened 2026-04-30
feat(relayer): per-bot broadcast observability + identify handshake

llamalm:
## Why

The relayer's auction.started fanout is fire-and-forget. The `bots=N` log line in `escrow.ts` only counts sockets that looked OPEN when the for loop started — it does not prove the relayer attempted delivery to a specific client, that the kernel buffer accepted the bytes, or that the bot's handler ran. As a result, when a funded market-maker silently misses an auction (incident: auction `e80a3944...` priced by the estimator only, with no live bid), the existing logs cannot answer:

- Was that bot connected at the time?
- Which socket was it?
- Did the relayer attempt to send to that socket?
- Did the send return success?
- Did the bot ever process the message?

This PR closes that gap. It is observability-only — no protocol change for clients that don't opt in.

Confirmed Railway state: relayer is `numReplicas: 1` (single region us-east4-eqdc4a). Distributed-fanout / split-brain is not on the table for the current incident, so this PR focuses on per-client tracing inside one process.

### Identity layers

Three identity fields stack so a bot is identifiable across reconnects, restarts, and parallel deployments:

| Field            | Source                                  | Changes on               |
|------------------|------------------------------------------|--------------------------|
| `clientId`       | relayer (`crypto.randomUUID()` on connect) | every WS reconnect       |
| `instanceId`     | client (per-process UUID)               | restart / redeploy       |
| `service`        | client (logical name)                   | never                    |
| `serviceInstance`| client (`RAILWAY_SERVICE_NAME`)         | redeploy to a new svc    |
| `replicaId`      | client (`RAILWAY_REPLICA_ID`)           | scale events             |
| `deploymentId`   | client (`RAILWAY_DEPLOYMENT_ID`)        | every deploy             |
| `version`        | client (`GIT_COMMIT_SHA`)               | every deploy             |

Two `auction-bidder` connections with different `instanceId` ⇒ two parallel processes (alertable). Same `instanceId`, new `clientId` ⇒ in-process reconnect. New `instanceId` ⇒ restart.

### SDK

- `IdentifyPayload` and `AuctionReceivedPayload` types added.
- `ClientToServerMessage` extended with `'identify'` and `'auction.received'`. Backwards-compatible: clients that never send identify are treated as `service: 'anonymous'`.

### Relayer

- `ClientConnection` carries mutable `service`/`instanceId`/`chainId`. `send()` now returns `boolean` so callers can count failures instead of swallowing them.
- `wsTransport` returns `false` when the socket is non-OPEN or `ws.send` throws; both paths log `clientId` and `service`.
- `subscriptions.broadcastRaw` uses the new bool to evict dead clients (was previously a try/catch on `void send`).
- `ws.ts` logs `client.connected clientId=... ip=...` and `client.closed clientId=... service=... instance=...` so the lifecycle is correlatable.
- `handlers/escrow.ts`:
  - Per-client broadcast log: `auction.broadcast.send auctionId=<8> clientId=<8> service=<name> instance=<8> ok=true|false`
  - Summary: `auction.broadcast.done auctionId=<8> attempted=N sent=K failed=F skippedClosed=O`
  - `handleIdentify` sets `client.service`/`instanceId`/`chainId` and logs `client.identified ...` once per identify message.
  - `handleAuctionReceived` logs `auction.client_ack ...` — proves end-to-end delivery (handler ran), not just that bytes were buffered.
- New Prometheus counters: `relayer_auction_broadcast_sends_total{service, ok}`, `relayer_auction_received_acks_total{service}`, `relayer_clients_identified_total{service}`.

## How this debugs the original incident

For auction `e80a3944...`, after deploying both this PR and the matching vault-bot identify PR you can answer:

- **Was `auction-bidder` connected?** Search for `client.identified ... service=auction-bidder` before the auction timestamp.
- **Did the relayer try to deliver?** `grep auction.broadcast.send auctionId=e80a3944` — one line per attempted recipient with service+ok.
- **Did the bot's handler run?** `grep auction.client_ack auctionId=e80a3944 ... service=auction-bidder`.

Diff between `attempted` and `sent` ⇒ `ws.send` failed. `sent=true` but no client_ack ⇒ bot received bytes but its handler didn't run (kernel buffer, pong-death, in-flight close, JS exception before handler).

## Deploy order

This PR ships first. Without it, vault-bot's `identify` and `auction.received` messages would be classified as unhandled by the older relayer's escrow type guard and contribute to the per-IP invalid-message penalty.

## Test plan

- [x] `pnpm --filter @sapience/relayer test` — 329/329 passing locally
- [x] `pnpm --filter @sapience/relayer build` — clean
- [ ] After merge + deploy: confirm new log lines appear in Railway and that pre-existing bots without identify still log as `service=anonymous`
- [ ] After vault-bot PR merges + deploys: verify `client.identified` for `auction-bidder` and `estimator`, and that every `auction.broadcast.send` to `auction-bidder` is followed by an `auction.client_ack`

## Out of scope

- No durable auction event log / "give me auctions I missed" replay (still 60s in-memory per `escrowRegistry.ts:18-20`).
- No multi-replica fanout. Current relayer is single-replica per Railway config; if that ever changes we'll need shared pubsub before this observability is sufficient.


packages/relayer/src/handlers/escrow.ts
```diff
@@ -120,22 +131,60 @@ export async function handleAuctionStart(
...
+  // Broadcast auction.started with auction details to all connected clients.
+  // Per-client logging here is intentional: it's the only ground truth we
+  // have for "did the relayer attempt to deliver this auction to this bot."
+  // Without it, a silent miss (auction received, but bot never saw it) is
+  // indistinguishable from a relayer-side failure.
   const details = getEscrowAuctionDetails(auctionId);
   if (details) {
     const broadcastMsg = { type: 'auction.started', payload: details };
-    let botCount = 0;
+    const auctionShort = auctionId.slice(0, 8);
```
ukitta555: use short() here

packages/relayer/src/handlers/escrow.ts
```diff
@@ -120,22 +131,60 @@ export async function handleAuctionStart(
...
+          `[Relayer] auction.broadcast.send auctionId=${auctionShort} clientId=${short(
+            c.id
+          )} service=${c.service} instance=${short(c.instanceId)} ok=true`
+        );
+      } else {
+        failed++;
+        console.warn(
+          `[Relayer] auction.broadcast.send auctionId=${auctionShort} clientId=${short(
+            c.id
+          )} service=${c.service} instance=${short(c.instanceId)} ok=false`
```
ukitta555: put "ok=false"  first so that it is easier to search for in the logs? nit

packages/relayer/src/handlers/escrow.ts
```diff
@@ -290,3 +339,76 @@ export async function handleBidSubmit(
...
+ * accepted the bytes.
+ */
+export function handleAuctionReceived(
+  client: ClientConnection,
+  payload: AuctionReceivedPayload | undefined
+): void {
+  if (!payload || typeof payload.auctionId !== 'string') return;
+  auctionReceivedAcks.inc({ service: client.service });
+  console.log(
+    `[Relayer] auction.client_ack auctionId=${payload.auctionId.slice(
```
ukitta555: short()

packages/relayer/src/handlers/escrow.ts
```diff
@@ -290,3 +339,76 @@ export async function handleBidSubmit(
...
+  const chainId =
+    typeof payload.chainId === 'number' && Number.isFinite(payload.chainId)
+      ? payload.chainId
+      : undefined;
+
+  client.service = service;
+  client.instanceId = instanceId;
+  client.chainId = chainId;
+
+  clientsIdentified.inc({ service });
```
nostradamus-agent: This uses client-supplied `service` directly as a Prometheus label (`clientsIdentified`, plus broadcast/ack labels after `client.service` is set). A public WS client can repeatedly identify with random service names and create unbounded time series. Please bucket this through an allowlist/enum (`unknown`/`anonymous` for everything else) before it reaches metrics.

packages/relayer/src/handlers/escrow.ts
```diff
@@ -290,3 +339,76 @@ export async function handleBidSubmit(
...
+ * handler ran and saw the auction). Without this ack, a successful `ws.send`
+ * is the strongest signal we have, and it only proves the kernel buffer
+ * accepted the bytes.
+ */
+export function handleAuctionReceived(
+  client: ClientConnection,
+  payload: AuctionReceivedPayload | undefined
+): void {
+  if (!payload || typeof payload.auctionId !== 'string') return;
+  auctionReceivedAcks.inc({ service: client.service });
```
nostradamus-agent: `auction.received` is counted for any client-provided `auctionId`, with no check that the auction exists or that this connection was actually sent that auction. That makes the metric spoofable, especially combined with arbitrary `identify`. Track recent sent auction IDs per connection and only count matching acks, or rename/downgrade this so dashboards don't treat it as proof of delivery.

nostradamus-agent (changes requested):
Found two observability correctness/security blockers in the new relayer handshake/ack path.

The local relayer package checks are green (`build:lib`, relayer `type-check`, `lint`, and full relayer tests), so this is not a compile/test issue — it is the metric trust boundary. The relayer is accepting unauthenticated WebSocket client metadata/acks and then using them as low-cardinality, proof-like Prometheus signals.

Suggested shape: normalize `service` through a small allowlist/bucket before labels, and only count `auction.received` when it matches an auction this connection was actually sent (or explicitly downgrade the metric/docs to "client claimed ack", not proof of delivery).

ukitta555 (approved):

llamalm:
Both blockers addressed in 87419ef.

**1. Bounded service metric label.** New `KNOWN_SERVICE_LABELS` allowlist in `metrics.ts` (`auction-bidder`, `estimator`, `secondary-bidder`, `vault-quoter`, `anonymous`) plus a `serviceLabel(raw)` helper that collapses anything else to `'unknown'`. Used at every `.inc({ service: ... })` site — `auctionBroadcastSends`, `auctionReceivedAcks`, `clientsIdentified`. The raw client-supplied string is still kept in `client.service` for log lines (sanitized of control chars to prevent log injection), but it can no longer materialize as a Prometheus label. Adding a new service requires a code change here, which is the right tradeoff for label cardinality.

**2. Anti-spoof for `auction.received`.** New `broadcastLedger.ts` keeps a `Map<auctionId, Map<clientId, ts>>` of which clients the relayer attempted to deliver each auction to. We populate it inside the broadcast loop in `handleAuctionStart`, and only when `c.send(...)` returns `true`. `handleAuctionReceived` now gates on `wasBroadcastTo(auctionId, client.id)`:

- Hit → counts as a real ack: `auctionReceivedAcks{service}` +1, `auction.client_ack ...` log line emitted.
- Miss → `auctionReceivedAckRejected{reason='not_recipient'}` +1, warning log, no ack counted.
- Malformed payload → `auctionReceivedAckRejected{reason='invalid_payload'}` +1.

Memory bound: TTL 90s (slightly over the 60s auction TTL to tolerate late acks) with a 30s sweep. Process-local; same scope as the existing in-memory escrow registry.

After this, `auction.client_ack` is real proof-of-delivery, not a self-report — a client cannot fabricate an ack for an auction it wasn't a recipient of, and a client whose `c.send` returned `false` cannot ack either (we only record successful sends).

**New tests** (340/340 passing):
- `serviceLabel allowlist` — known returns raw, unknown collapses, empty/undefined → `'unknown'`
- `handleIdentify` — sets fields, sanitizes control chars, uses bounded label for metric
- `handleAuctionReceived (anti-spoof)` — rejects unknown auctionId, malformed payload, and a client B trying to ack an auction broadcast only to client A
- `handleAuctionStart > records broadcast recipients only when send succeeded` — failed-send recipients can't ack
- `handleAuctionStart > uses bounded service label in broadcast metric` — unknown `c.service` becomes `'unknown'` in the counter

llamalm:
Follow-up for the cardinality fix — turns out a single allowlisted `service` per logical bot kind (`auction-bidder`, `vault-quoter`, …) collapses what we'd actually want to break out: today there's already a main vault auction-bidder + a pyth vault auction-bidder running side-by-side, and a third "experimental" vault is on the horizon. With only `service` as a label, all three of those would land on the same `auction-bidder` series.

Added an orthogonal `variant` label so we can distinguish them without going back to free-form labels:

- SDK: `IdentifyPayload` gets optional `variant?: string`. Backwards-compatible — clients without it default to `'default'` server-side.
- Relayer: `KNOWN_VARIANT_LABELS = ['default', 'pyth', 'experimental']` + `variantLabel()` mirroring the existing `serviceLabel()`. Anything outside the set collapses to `'unknown'`. Empty/undefined → `'default'` so the legacy series stays intact.
- Counters get the new label: `auctionBroadcastSends{service, variant, ok}`, `auctionReceivedAcks{service, variant}`, `clientsIdentified{service, variant}`.
- Cardinality stays bounded: 6 service labels × 4 variant labels × 2 ok values = 48 series for the busiest counter. Adding a new variant is the same one-line allowlist edit + redeploy as adding a service.

Idiomatic Prometheus shape too — `sum by (service) (rate(relayer_auction_broadcast_sends_total[5m]))` rolls up across vaults, `sum by (service, variant) (...)` breaks them out.

Tests at 345/345. New cases:
- `variantLabel allowlist` — known passes through, empty/undefined → `'default'`, unknown → `'unknown'`
- `handleIdentify` — variant set/sanitized; default applied when omitted; bounded for metric
- `handleAuctionStart > breaks out broadcast metric per (service, variant) pair` — main bidder + pyth bidder produce separate series
- `handleAuctionStart > uses bounded service+variant labels in broadcast metric` — rogue values collapse together

Vault-bot side (sibling PR sapiencexyz/vault-bot#224): the identify helper now derives `variant` automatically — explicit param wins, then `BOT_VARIANT` env, then `PRICING_STRATEGY=pyth` ⇒ `'pyth'`, else `'default'`. The pyth wrappers in `entries/` already set `PRICING_STRATEGY=pyth`, so they'll identify as `variant=pyth` with no further code changes; main bots stay `variant=default`.

packages/relayer/src/handlers/escrow.ts
llamalm: Done in 9efe926.

packages/relayer/src/handlers/escrow.ts
llamalm: Good call. Applied to both the success and failure log lines for consistency in 9efe926, so `ok=` is now the first key=value. `grep 'ok=false'` lights up the failures and the warn-vs-info distinction stays in the log level.

packages/relayer/src/handlers/escrow.ts
llamalm: Done in 9efe926 — both the `auction.client_ack` and the `auction.received rejected` lines use `short()` now.

nostradamus-agent (approved):
Re-reviewed the current head. My previous blockers are addressed: metric labels are now bounded through `serviceLabel`/`variantLabel`, identity strings are sanitized before logs, and `auction.received` is gated by a per-client broadcast ledger instead of accepting arbitrary auction IDs.

Local validation I ran:
- `pnpm --filter @sapience/relayer run type-check`
- `pnpm --filter @sapience/relayer run lint`
- `pnpm --filter @sapience/relayer run test` — 17 files / 345 tests passed

Tiny caveat, not blocking: service identity is still self-reported by the WS client, so dashboards should read this as “a connection identifying as service X received the auction”, not cryptographic bot authentication. For the observability goal here, this is good enough.

PR #1662, opened 2026-04-30
fix(app): trim 1s from displayed bid deadlines for inclusion latency

noahlitvin:
## Summary
- Introduce `DEADLINE_DISPLAY_BUFFER_SECS = 1` plus `effectiveDeadlineMs` / `remainingMsToDeadline` / `isBidExpired` helpers in `packages/app/src/lib/auction/bidExpiry.ts`, and apply them everywhere the app compares a counterparty bid's deadline against `nowMs` for user-facing purposes.
- The submitted on-chain `counterpartyDeadline` value is unchanged — we only nudge the visible window in by 1s so a user who acts at the last moment still leaves room for the UserOp bundler / inclusion latency. Without this buffer the on-chain deadline check silently returns `false` and the protocol surfaces the failure as `InvalidCounterpartySignature`.

### Touched
- `BidDisplay`: countdown text + `isBidExpired` (drives the submit-button gate)
- `CreatePositionForm/index`: submit-click defense in depth
- `PositionForm`, `selectBestBid`: best-bid selection
- `AuctionRequestInfo`: terminal countdown + top/winning-bid filters
- `MarketPredictionRequest`, `AuctionBidsChart`: auxiliary displays
- `usePreprocessedBids`, `useValidatedBids`, `useComboQuotes`, `useSingleConditionAuction`: bid-list filters that drive what the UI ends up showing

### Drive-bys
The precommit hook surfaced two pre-existing lint warnings on files I touched, both fixed:
- `AuctionRequestInfo`: import order (third-party before local)
- `useSingleConditionAuction`: switched the direct `selectedTakerAddress` read inside `triggerQuoteRequest` to `selectedTakerAddressRef.current`, matching the ref pattern the same callback already uses on its early-return guard

## Test plan
- [x] `pnpm --filter @sapience/app run test` (528 / 528 pass)
- [ ] Open a market, request a bid, watch the countdown — confirm "Xs" displays one second short of the on-chain `counterpartyDeadline` and the SUBMIT button disables one second before it would have on `staging`.
- [ ] In the terminal, watch a bid in `AuctionRequestInfo` tick down to 0 and confirm it disappears from the best-bid table 1s earlier than the underlying deadline.

### DIFF
--- a/packages/app/src/components/terminal/AuctionRequestInfo.tsx
+++ b/packages/app/src/components/terminal/AuctionRequestInfo.tsx
@@ -4,9 +4,6 @@ import type React from 'react';
 import { useEffect, useMemo, useState } from 'react';
 import { formatDistanceToNowStrict } from 'date-fns';
 import { formatEther } from 'viem';
-import EnsAvatar from '~/components/shared/EnsAvatar';
-import { AddressDisplay } from '~/components/shared/AddressDisplay';
-import PlaceBidForm from '~/components/terminal/PlaceBidForm';
 import {
   Tooltip,
   TooltipContent,
@@ -14,8 +11,15 @@ import {
   TooltipTrigger,
 } from '@sapience/ui/components/ui/tooltip';
 import { HelpCircle } from 'lucide-react';
+import EnsAvatar from '~/components/shared/EnsAvatar';
+import { AddressDisplay } from '~/components/shared/AddressDisplay';
+import PlaceBidForm from '~/components/terminal/PlaceBidForm';
 import ExpiresInLabel from '~/components/shared/ExpiresInLabel';
 import PercentChance from '~/components/shared/PercentChance';
+import {
+  effectiveDeadlineMs,
+  remainingMsToDeadline,
+} from '~/lib/auction/bidExpiry';
 
 type SubmitData = {
   amount: string;
@@ -66,7 +70,7 @@ const BestBid: React.FC<BestBidProps> = ({
         const deadlineSec = Number(b?.counterpartyDeadline || 0);
         const ms =
           Number.isFinite(deadlineSec) && deadlineSec > 0
-            ? deadlineSec * 1000
+            ? effectiveDeadlineMs(deadlineSec)
             : Number.POSITIVE_INFINITY;
         if (ms > now) return b;
       }
@@ -90,9 +94,9 @@ const BestBid: React.FC<BestBidProps> = ({
                 const secondsRemaining = (() => {
                   if (!Number.isFinite(deadlineSec) || deadlineSec <= 0)
                     return null;
-                  const ms = deadlineSec * 1000;
-                  const diff = Math.max(0, Math.round((ms - now) / 1000));
-                  return diff;
+                  return Math.round(
+                    remainingMsToDeadline(deadlineSec, now) / 1000
+                  );
                 })();
                 const payoutStr = (() => {
                   try {
@@ -275,7 +279,7 @@ const AuctionRequestInfo: React.FC<Props> = ({
     } catch {
       return null;
     }
-  }, [maxEndTimeSec, now]);
+  }, [maxEndTimeSec]);
 
   const maxRemainingExpirySeconds = useMemo(() => {
     try {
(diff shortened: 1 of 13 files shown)


packages/app/src/components/terminal/AuctionRequestInfo.tsx
```diff
@@ -275,7 +279,7 @@ const AuctionRequestInfo: React.FC<Props> = ({
     } catch {
       return null;
     }
-  }, [maxEndTimeSec, now]);
+  }, [maxEndTimeSec]);
```
nostradamus-agent: This dependency change freezes the `Maximum Duration` label. `formatDistanceToNowStrict()` reads the current time internally, and the component's `now` interval is what was forcing this memo to recompute once per second. With `now` removed, the component still rerenders, but `maxDurationLabel` stays cached until `maxEndTimeSec` changes. Please keep `now` in the dependency list (or compute from `maxRemainingExpirySeconds`) so the label actually counts down.

nostradamus-agent (changes requested):
🔮 Review pass found one small regression to fix before merge.

The buffered deadline helper itself looks clean, and the new unit coverage for `bidExpiry` passed locally. But `AuctionRequestInfo` accidentally stopped recomputing the displayed max-duration label on the 1s tick, so that terminal label will freeze after mount.

Verification:
- `git diff --check` ✅
- `pnpm exec vitest run src/lib/auction/__tests__/bidExpiry.test.ts --pool=forks --poolOptions.forks.singleFork --reporter=verbose` ✅ (8 tests)
- `pnpm --filter @sapience/app run type-check` ⚠️ local worktree dependency links are broken in this sandbox, so this produced missing-module noise unrelated to the PR bytes; relying on CI/author run for full app typecheck.

nostradamus-agent (approved):
Reviewed the buffered bid-deadline changes. The main flow looks sane: the helper keeps the on-chain `counterpartyDeadline` intact, applies the 1s client-side cutoff consistently across selection/submit/display paths touched by the PR, and adds focused unit coverage.

Verification run locally in an isolated worktree:
- `git diff --check 2070ae600c1badb538aa4ae022405062df348de3...HEAD`
- `pnpm exec vitest run src/lib/auction/__tests__/bidExpiry.test.ts src/lib/auction/__tests__/selectBestBid.test.ts --pool=forks --poolOptions.forks.singleFork --reporter=verbose` — 16 tests passed
- `pnpm exec vitest run src/hooks/auction/__tests__/useValidatedBids.test.ts src/hooks/auction/__tests__/usePreprocessedBids.test.ts --pool=forks --poolOptions.forks.singleFork --reporter=verbose` — 27 tests passed

One non-blocking follow-up: `packages/app/src/lib/utils/positionFormUtils.ts:getBestDisplayBid()` still does the raw `counterpartyDeadline * 1000 > nowMs` comparison for OG-preview payout selection. It won’t bypass the submit gate, but if we want the display-buffer rule to be truly universal for user-facing quote display, that helper should probably use `effectiveDeadlineMs()` too.

Nit from the sandbox: full app lint could not run cleanly from the lightweight worktree because package-local dependency symlinks are stale/missing (`@eslint/js`, then `eslint-plugin-react`); I’m not treating that as a PR failure. Vercel statuses on the head SHA are green.

nostradamus-agent (approved):
Reviewed the buffered bid-deadline changes. The main flow looks sane: the helper keeps the on-chain `counterpartyDeadline` intact, applies the 1s client-side cutoff consistently across selection/submit/display paths touched by the PR, and adds focused unit coverage.

Verification run locally in an isolated worktree:
- `git diff --check 2070ae600c1badb538aa4ae022405062df348de3...HEAD`
- `pnpm exec vitest run src/lib/auction/__tests__/bidExpiry.test.ts src/lib/auction/__tests__/selectBestBid.test.ts --pool=forks --poolOptions.forks.singleFork --reporter=verbose` — 16 tests passed
- `pnpm exec vitest run src/hooks/auction/__tests__/useValidatedBids.test.ts src/hooks/auction/__tests__/usePreprocessedBids.test.ts --pool=forks --poolOptions.forks.singleFork --reporter=verbose` — 27 tests passed

One non-blocking follow-up: `packages/app/src/lib/utils/positionFormUtils.ts:getBestDisplayBid()` still does the raw `counterpartyDeadline * 1000 > nowMs` comparison for OG-preview payout selection. It won’t bypass the submit gate, but if we want the display-buffer rule to be truly universal for user-facing quote display, that helper should probably use `effectiveDeadlineMs()` too.

Nit from the sandbox: full app lint could not run cleanly from the lightweight worktree because package-local dependency symlinks are stale/missing (`@eslint/js`, then `eslint-plugin-react`); I’m not treating that as a PR failure. Vercel statuses on the head SHA are green.

PR #1664, opened 2026-04-30
refactor(admin): consolidate referrals analytics into GraphQL, drop reindex tools

noahlitvin:
## Summary

- Delete the unused admin Reindex/RFQ/ConditionGroups UI plus the matching reindex worker jobs, REST routes, and `start:reindex-*` npm scripts.
- Move public referral analytics from three REST endpoints into a single GraphQL query: `Query.referralCodes(id, limit, cursor)` returning `ReferralCodesPage` with `claimCount`, `totalVolume`, `totalPositions`, and a paginated `claimants` field. All sums are computed in Postgres via `$queryRaw`, so we no longer load the full `position` table over the wire.
- Strip `codeHash` from the SDL and every response. `keccak256` of short user-chosen codes is GPU-brute-forceable, so a leaked hash is credential-equivalent (worst case: volume mis-attribution). Identity for clients is the integer `id`. Schema validation now rejects any client trying to select `codeHash`.
- Restore the EOA → smart-account lookup dialog on the admin page.
- Regenerate SDK GraphQL types from the new schema.

## Removed surface

| | |
|---|---|
| REST | `/referrals/codes`, `/referrals/codes/:id`, `/referrals/codes/:id/analytics` |
| Worker jobs | `reindexAccuracy`, `reindexConditionSettled`, `reindexTransfers`, `reindexCollateralTransfers` |
| Admin REST routes | `/admin/reindex/*` |
| FE components | `RFQTab`, `ConditionGroupsTab`, `ConditionStatusBadges`, `Reindex*Form` |
| GraphQL schema | `ReferralCode.codeHash`, `ReferralCode.claimedBy`, `ReferralCodeCount`, the `codeHash` filter/sort inputs |
| npm scripts | `start:reindex-accuracy`, `start:reindex-condition-settled`, `start:reindex-transfers`, `start:reindex-collateral-transfers` |

## Test plan

- [x] `pnpm --filter @sapience/api run lint` (clean)
- [x] `pnpm --filter @sapience/api run type-check` (clean)
- [x] `pnpm --filter @sapience/api run test` — 419/419 pass
- [x] `pnpm --filter @sapience/app run lint` + `type-check` (clean)
- [x] New `Query.referralCodes` resolver test (10 cases): pagination, SQL-aggregation invariant, `codeHash` absence, id-mode, fallback stats
- [x] Live: `POST /graphql` with `referralCodes(limit:3)` returns real data with inline analytics
- [x] Live: `referralCodes(id:48){items{claimants(limit:5){items{...}}}}` returns the per-code claimant breakdown
- [x] Live: `query{referralCodes(limit:1){items{codeHash}}}` is rejected with `Cannot query field "codeHash"`
- [x] Live: deleted REST routes return 404
- [x] Live: `/admin` page renders with new Created column + restored Smart Account Lookup


packages/api/src/graphql/sdl/resolvers/queries/referrals.ts
```diff
@@ -0,0 +1,132 @@
...
+      totalVolume: row.total_volume ?? '0',
+      totalPositions: Number(row.total_positions ?? 0),
+    });
+  }
+  return result;
+}
+
+export const referralCodes: NonNullable<
+  QueryResolvers['referralCodes']
+> = async (_parent, args) => {
```
nostradamus-agent, line 95: 🔴 This resolver is now the replacement for the admin-protected referral list/analytics REST endpoints, but it has no admin auth check and GraphQL context only carries `prisma`. Anyone who can hit `/graphql` can enumerate referral codes plus `createdBy`, `claimCount`, `totalVolume`, and `totalPositions`. Please either add an admin-authenticated GraphQL path/context check or keep this analytics surface behind the existing admin REST middleware.

packages/app/src/hooks/referrals/useAdminReferralCodes.ts
```diff
@@ -88,31 +92,100 @@ function useReferralAdminFetch() {
...
-  const referralAdminFetch = useReferralAdminFetch();
   return useQuery<AdminReferralCodeRow[]>({
     queryKey: ADMIN_CODES_QUERY_KEY,
-    queryFn: () =>
-      referralAdminFetch<AdminReferralCodeRow[]>(
-        '/referrals/admin/codes',
-        'GET'
-      ),
+    queryFn: async () => {
+      const data = await graphqlRequest<{
```
nostradamus-agent: This read path switched from signed admin REST requests to the generic public `graphqlRequest`, so it sends no `x-admin-signature` and ignores the admin endpoint setting. If the server fixes the auth hole, this hook will fail; until then, the admin UI is reading sensitive analytics unauthenticated.

packages/app/src/components/admin/ReferralCodesTab.tsx
```diff
@@ -235,34 +237,32 @@ const ReferralCodesTab = ({
   const columns: ColumnDef<ReferralCodeRow>[] = useMemo(
     () => [
       {
-        header: 'Code Hash',
-        accessorKey: 'codeHash',
+        header: 'Created',
```
nostradamus-agent, line 240: Non-blocking UI regression: replacing `Code Hash` with another `Created` column leaves the table with two `Created` columns and no visible non-sensitive row identifier. Since `codeHash` is intentionally gone, show `id` or another safe identifier here instead.

nostradamus-agent (changes requested):
## Nostradamus review

Requesting changes — the GraphQL migration moves admin referral analytics onto the public GraphQL surface without an auth boundary. That exposes referral-code inventory and claimant wallet/volume analytics that were previously behind `adminAuth`.

Also: the PR says `codeHash` is credential-equivalent and strips it from GraphQL, but the remaining REST create/update responses still return `codeHash` (`POST /referrals/code`, `POST /referrals/admin/codes`, `PUT /referrals/admin/codes/:id`). If the threat model is right — and I think it is — those response paths need the same treatment.

Verification I ran:
- `git diff --check` ✅
- stale deleted-admin-component reference search ✅
- `pnpm exec vitest run src/graphql/sdl/resolvers/queries/referrals.test.ts --pool=forks --poolOptions.forks.singleFork --reporter=verbose` ✅ (10 tests)
- app targeted test attempted, but sandbox dependency symlinks are busted (`lucide-react` resolution), so no trustworthy app test signal locally.

nostradamus-agent:
Clarifying based on Noah's note: if `referralCodes` is intentionally public, then the auth-boundary part of my review is a product/API-surface decision, not an accidental bug. I’m fine downgrading that piece if the team explicitly wants referral inventory + aggregate analytics public.

What I still think should block before merge:

1. `codeHash` is still returned by remaining REST create/update responses. The PR's threat model says this value is credential-equivalent, so GraphQL removing it is only half the fix.
2. Please make the publicness explicit in code/docs/tests: add a resolver/schema test showing `referralCodes` is intentionally queryable without admin auth, and ideally call out that claimant wallet + volume breakdown is public too. That avoids the next reviewer rediscovering this and thinking it’s an auth regression.

Non-blocking cleanup still stands: the app table has duplicate `Created` columns and no safe identifier; showing `id` would be clearer.

nostradamus-agent (approved):
Re-reviewed after Noah clarified the intended public read surface.

I retract the auth/codeHash blockers:
- `referralCodes` being public is intentional.
- The remaining `codeHash` responses are create/update paths where the caller is already credentialed / supplying the plaintext or admin signature. Claim does not return it.

Only leftover is non-blocking UI polish: duplicate `Created` columns / no visible safe id in the referral-code table.

Local verification from previous pass:
- `git diff --check` ✅
- stale deleted-admin-component search ✅
- API referral resolver test ✅

PR #1667, opened 2026-05-01
Profile + feed infinite scroll, server-truth pagination, and shared loading UI

noahlitvin:
## Summary

Adds infinite scroll to the **Profile → Positions** tab, fixes a class of pagination bugs the rollout exposed, eliminates a graph-of-waterfalls that was making the page feel slow, and consolidates the loading / empty / load-more UI across the three profile-tab tables (Positions / Forecasts / Activity) plus the auction `/feed` page.

It's broader than the title suggests because the original feature surfaced several pre-existing layout, pagination, and complexity issues — fixing them ended up touching the API, SDK, FE hooks, the layout root, and three profile-tab tables.

### Profile → Positions infinite scroll
- `usePositionBalances` / `usePositionBalancesByConditionId` switched from one-shot `useQuery` to `useInfiniteQuery`, paged at 15 (was an unpaginated implicit cap of 50).
- `PositionsTable` adds an `IntersectionObserver`-driven sentinel via the shared `useInfiniteScroll` hook.

### Server-truth pagination (the blocker)
The positions resolver synthesizes per-row events from raw `Position` rows. Some raw pages can emit zero items (zero-balance unresolved positions with no sells), so the naïve "stop on `lastPage.length === 0`" rule on the client could halt pagination before later visible rows.

- New `positionsPage(...): { items: [Position!]!, hasMore: Boolean! }` GraphQL query.
- Resolver fetches `take + 1` raw rows and reports `hasMore` from raw count, so empty-synthesized pages don't terminate scroll early.
- FE hooks read `lastPage.hasMore` instead of inferring from `items.length`.

### Drop the conditions waterfall
The Ends column / category needed `condition` data per pick, which used to require a follow-up `useConditionsByIds` round trip — visible 1-2s delay between data and full row render.

- Inline `Pick.condition` field via SDL + per-request `ctx.pickConditions` Map cache; resolvers pre-batch a single `condition.findMany` per request.
- Shared `preloadPickConditions` helper used by `escrow.predictions`, `escrow.prediction`, `escrow.pickConfigurations`, `escrow.pickConfiguration`, `escrow.positions`, and `accountActivity`.
- FE: `usePositions`, `useAccountActivity`, `useRecentCombos`, `useEnrichedTokens` build their `conditionsMap` from inline data; the secondary `useConditionsByIds` call is gone.

### Resolver perf
- Push the holder filter into the `Picks.predictions` include — the relation pulls every user's prediction otherwise. For popular pickConfigs this is a 10x+ payload reduction. Local benchmark: warm-cache `positions(take: 50)` dropped ~4.5s → ~1.6s.
- Run `preloadPickConditions` and `secondaryTrade.findMany` in parallel via `Promise.all` rather than sequentially.

### GraphQL ceilings
- `GRAPHQL_MAX_COMPLEXITY` 10000 → 15000 (the inline `Pick.condition` pushed legit queries from ~8700 to ~11000; hardcoded ceiling left no headroom).
- `depthLimit` 5 → 6 (`accountActivity → prediction → pickConfig → picks → condition → category → slug` naturally needs 6).

### Layout / loading UI consistency (Positions / Forecasts / Activity / Feed)
The rollout made several pre-existing layout fragilities visible:

- **Fixed footer overlay**: `<Footer>` is `sm:fixed; bottom: 0; ~33px` and the layout reserved no space for it. Page content slid under it. Fix: `sm:pb-[33px]` on the layout root.
- **Hardcoded viewport offsets** for the loading/empty box — replaced with a flex-based fill (`fill?: boolean` on the three tables, `flex-1` cascades through the page wrapper / bordered container / TabsContent / inner wrapper).
- **`/feed` route wrapper** was `<div class="relative min-h-screen">` (display:block) and broke the flex cascade — switched to `flex flex-1 flex-col`.
- **Radix TabsContent gotcha**: a plain `flex` class on inactive panels beat the UA `[hidden] { display: none }` rule (equal specificity, cascade order), so all three panels split the bordered container's height via `flex-1` and the active one only got ~1/3. Gated the flex utilities on `data-[state=active]:`.
- **`ProfilePageContent` opacity-0 gate**: the tabs region was hidden behind `opacity-0` until forecasts finished loading, which obscured each tab's internal loader during slow positions queries — gate dropped.
- **`ForecastsTable` wrapping `<div>`** clamped the flex chain to intrinsic height — wrap now propagates `flex-1 flex flex-col` when `fill` is set.
- **Triple-stacked bottom padding** (`mb-8` + page-wrapper `pb-8` + layout `pb-[33px]` = 97px below the table) — dropped the redundant page-wrapper `pb`. Visible bottom gap above the fixed footer now matches the side `px-3 md:px-6 lg:px-8` (12 / 24 / 32px) at each breakpoint.

### Shared loading UI
- New `<TableLoadingState>` (small inline `<Loader>` + `font-mono uppercase` "Loading …" message, optionally fills the viewport via flex-1) and `<InfiniteScrollFooter>` (load-more sentinel with consistent loading/idle messaging plus `border-t`). All three tables use them.
- `EmptyTabState` gains a `fill` prop with the same flex-1 behaviour.
- WON/LOST badge now appears in the Ends column for resolved positions and for unresolved-but-closed-via-secondary positions where PnL implies the outcome.

### Default page sizes / smoothness
- Positions default page size 50 → 15 so initial paint is fast on infinite scroll.
- Activity pagination similarly tightened (separate commit on the branch).

### Slim chart query
`PREDICTIONS_BY_CONDITION_QUERY` powering the question-page scatter only needs ~6 fields and `pickConfig.picks.{conditionId,conditionResolver,predictedOutcome}`. Trimmed the query — keeps `take: 100` under the 15k complexity ceiling.

## Tests

- 8 new resolver tests in `escrow.test.ts` covering: predictions filter is pushed into the include with holder-pinned, preserved when not, parallel preload + trades fetch, empty-rows fast path, `take + 1` raw fetch, `hasMore` true/false, and the regression case (zero synthesized items but more raw rows remaining).
- 5 new hook tests in `usePositions.test.ts` covering server-truth pagination shape: empty-page stop, partial-page continue (synthesized vs raw count mismatch), parameter pass-through, disabled-when-no-holder, and the conditionId variant.
- New `preloadPickConditions.test.ts`.
- Updated `useAccountActivity.test.ts` for the new query shape.
- Contract snapshots regenerated (introspection + schema SDL + `pickConfigurations` operation).

## CI status (local)
- `pnpm --filter @sapience/app run type-check` — clean
- `pnpm --filter @sapience/api run type-check` — clean
- `pnpm --filter @sapience/app run lint` — clean
- `pnpm --filter @sapience/api run lint` — clean
- `pnpm --filter @sapience/app run test` — 533 / 533 ✓
- `pnpm --filter @sapience/api run test` — 419 / 419 ✓
- `pnpm --filter @sapience/sdk run test` — 649 / 649 ✓
- API contract tests — 18 / 18 ✓

## Test plan
- [ ] On `/profile/<address>#positions`, scroll past the 15th row → next page loads automatically; eventually the "Scroll to load more" footer flips to nothing when exhausted.
- [ ] On `/profile/<address>#forecasts` and `#activity`, same infinite-scroll behaviour with the same loading/empty/load-more UI.
- [ ] Resolved positions show WON / LOST in the Ends column; unresolved-but-closed-via-secondary positions show WON / LOST when the PnL is decisive.
- [ ] `/feed` initial load shows "LOADING ACTIVITY…" filling the viewport (centered vertically); as the user scrolls, the load-more footer shows "Loading more activity…" with a border-top above it; bottom gap matches the side padding.
- [ ] Empty/error states fill the viewport on profile + feed but render compact (`py-12`) inside `PositionDialog` and the `/questions/:id` tabs.
- [ ] Filters on Positions table: status (active/won/lost) pushes `settled` to the server; search/value/date narrow already-loaded pages without re-fetching.
- [ ] Question-page scatter chart still renders; activity tab on `/questions/:id` works with the deeper `accountActivity → … → category` selection (previously hit the depth-5 limit and 400'd).


packages/app/src/hooks/graphql/usePositions.ts
```diff
@@ -381,79 +440,134 @@ export function usePredictions(params: {
...
+    queryKey: ['positionBalances', holder, chainId, settled, pageSize],
     enabled,
     staleTime: 30_000,
     gcTime: 5 * 60 * 1000,
     refetchOnWindowFocus: false,
     refetchOnReconnect: false,
-    queryFn: async () => {
+    initialPageParam: 0,
+    getNextPageParam: (lastPage: PositionBalance[], allPages) =>
+      lastPage.length < pageSize ? undefined : allPages.length * pageSize,
```
nostradamus-agent: ⚠️ This can terminate pagination early. `positions` paginates raw DB rows, then the resolver synthesizes/drops rows before returning. If a full raw page produces fewer than `pageSize` returned rows, `lastPage.length < pageSize` becomes true and the client never asks for the next raw page. We need explicit pageInfo/hasNextPage from the API, or server-side pagination after synthesis/filtering. Same issue in `usePositionBalancesByConditionId` below.

packages/app/src/components/positions/PositionsTable.tsx
```diff
@@ -808,6 +843,25 @@ export default function PositionsTable({
           </TableBody>
         </Table>
       </div>
+      {hasMore && (
+        <div
```
nostradamus-agent: ⚠️ This sentinel is unreachable when client-side filters produce zero matches on the currently loaded pages, because the component returns `No positions match your filters` above. With search/date/collateral filters still applied client-side, matches can exist on later pages but the user has no way to load them. Either keep the load-more sentinel visible while `hasMore` is true, or move those filters server-side before pagination.

nostradamus-agent (changes requested):
## 🔮 Nostradamus review

Requesting changes — the direction is good, but this is not merge-ready yet.

### Blockers

1. **API contract snapshots are stale, and CI is red because of it.** `test-api` is failing in the contract suite: `introspection.test.ts`, `schema.snapshot.test.ts`, and `operations/pickConfigurations.test.ts`. The PR adds `Pick.condition` to `packages/api/schema.graphql` / SDK query output, but does not update:
   - `packages/api/test/contract/__snapshots__/schema.sdl.graphql`
   - `packages/api/test/contract/__snapshots__/introspection.json`
   - `packages/api/test/contract/__snapshots__/operations/pickConfigurations.json`

2. **Positions infinite scroll can stop early.** The client decides `hasNextPage` from the synthesized rows returned by `positions`, but the server paginates raw `Position` rows before dropping/splitting into synthetic rows. A full raw page can return fewer than `pageSize` visible rows, so the client can conclude there is no next page while more raw rows exist.

3. **Filtered empty states can strand later matches.** Search/date/collateral filters are still client-side over loaded pages only. If the loaded page has no matches but later pages do, the component returns the empty state before rendering the load-more sentinel, so the user cannot scroll into the matching page.

### Verified
- `git diff --check` is clean.
- GitHub checks: app/sdk/relayer/market-keeper/protocol checks are green/skipped as expected; `test-api` is the failing gate above.

nostradamus-agent:
Follow-up performance pass, since Noah asked for optimization eyes:

### Perf issues worth fixing while this is open

1. **Activity infinite scroll is O(n²) over the wire/cache.**
   `packages/app/src/hooks/graphql/useAccountActivity.ts:204-236,280-281` grows `take` and always queries `skip: 0`, with `take` in the React Query key. Scrolling 20 → 40 → 60 rows downloads/cache-stores the whole prefix every time. Now that each row includes inline `pickConfig.picks.condition`, that gets expensive fast. This should be a real `useInfiniteQuery`: stable key, `pageParam` as `skip`, fetch only the next page, flatten pages.

2. **`condition.category` still N+1s server-side.**
   `packages/api/src/graphql/sdl/resolvers/preloadPickConditions.ts:34-36` batches `Condition` rows, but the GraphQL selection asks for `condition.category.slug`. `Condition.category` in `packages/api/src/graphql/sdl/resolvers/Condition.ts:16-22` falls back to `loadRelation`, so cached condition parents without `category` still do one category relation query per distinct condition. Include `category` in the preload or add a category DataLoader/request cache.

3. **Payload bloat from full inline condition hydration.**
   The app/SDK queries now pull `description`, settlement fields, estimated price, and category for every pick on every row (`usePositions.ts`, `useAccountActivity.ts`, `usePickConfigsByTokens.ts`, `sdk/queries/pickConfigurations.ts`). Removing the old client waterfall is good, but duplicating full condition objects across infinite-scroll rows is the new tax. Trim per-view fragments — table rows probably don’t need `description`; dialogs can fetch richer details lazily if needed.

4. **Client-side sorting/filtering does not scale with pagination.**
   `PositionsTable.tsx` sorts/filters the loaded subset only, so it is both not globally correct while `hasMore` is true and gets more expensive as pages accumulate. Longer-term, move sort/filter inputs into the API query; short-term, precompute sortable/filterable row metadata once per row instead of doing `BigInt`, `formatEther`, `Date`, and condition-result work inside sort/filter loops.

The first two are the ones I’d treat as blocking/perf-regression candidates. The other two are “don’t let this become the next table that melts profiles with real history.” 🔮

nostradamus-agent (approved):
## 🔮 Re-review

Approved. The earlier blockers are handled on the current head:

- contract snapshots regenerated; `test-api` is green
- positions pagination no longer stops on partial synthesized pages
- filtered empty states keep a sentinel while more pages exist
- activity pagination now uses `useInfiniteQuery`/`skip` pages instead of re-fetching a growing prefix
- `preloadPickConditions` includes `category`, avoiding the nested `condition.category` N+1

CI is green on the current head.

packages/app/src/hooks/graphql/usePositions.ts
```diff
@@ -381,79 +440,143 @@ export function usePredictions(params: {
...
-    queryFn: async () => {
+    initialPageParam: 0,
+    // The server paginates raw Position DB rows by (take, skip), then
+    // synthesizes per-row events (one Open + one synthetic SOLD per
+    // sell). The returned synthesized count therefore differs from the
+    // raw page size — using `lastPage.length < pageSize` would stop
+    // early on resolved-only pages and over-fetch on heavy-sell pages.
+    // Stop only when a page comes back empty: that's the unambiguous
+    // signal we've exhausted raw rows. Costs one trailing empty fetch.
+    getNextPageParam: (lastPage: PositionBalance[], allPages) =>
```
nostradamus-agent: 🔴 This is still not an unambiguous end-of-raw-rows signal. The resolver can return `[]` for a raw page that contains only zero-balance unresolved rows with no sells — `escrow.ts` explicitly drops those rows after the DB `take/skip`. In that case this hook stops even though later raw pages can contain visible positions. We need API pageInfo based on the raw fetch (`take + 1` / cursor), or the resolver should keep scanning raw pages until it finds visible rows or exhausts raw rows.

nostradamus-agent (changes requested):
## 🔮 Re-review after new commits

Requesting changes again — one pagination blocker remains.

The latest commit changes the positions hook to stop only when the resolver returns an empty synthesized page. That is safer than stopping on a partial page, but it is still not a server-truth signal. The `positions` resolver can return an empty synthesized page even when the raw DB query returned rows, because it explicitly drops zero-balance unresolved rows with no sells. If one raw page is made entirely of those dropped rows, the client stops and never reaches later visible positions.

Fix shape: the API needs to return explicit page metadata based on the raw fetch (`hasNextPage` via `take + 1`, a raw-row cursor, etc.), or the resolver needs to continue scanning raw pages until it either finds visible rows or exhausts raw rows. Then the hook should use that server truth instead of inferred synthesized length.

CI was still running when I reviewed; `git diff --check` is clean.

nostradamus-agent (approved):
Re-reviewed latest branch. The server-truth positions pagination shape fixes the earlier synthesized-empty-page blocker, and CI was red only because `positionsPage` was missing from generated SDK GraphQL types. I pushed `6fa543886` to regenerate `packages/sdk/types/graphql.ts`.

Verified locally: generated files are clean after `generate-types`, API compile passes, API/app targeted pagination tests pass, API tests pass. GitHub CI is green on the pushed head.

packages/api/src/graphql/startApolloServer.ts
```diff
@@ -22,13 +22,28 @@ export interface ApolloContext {
...
+   * referenced Condition up front and populate this map; the
+   * Pick.condition field resolver then returns rows from the cache
+   * without a per-pick round trip. Resolvers that don't pre-populate
+   * fall back to per-pick lookups.
+   *
+   * Stored on context (not as a request-scoped DataLoader) because we
+   * already do the batching ourselves before mapPickConfig and just
+   * need a place to stash the map for the field resolver to read.
+   */
+  pickConditions?: Map<string, unknown>;
```
ukitta555, line 38: this seems like cache; do we invalidate it at some point? do we need to invalidate it at all?

packages/api/src/graphql/sdl/resolvers/preloadPickConditions.ts
```diff
@@ -0,0 +1,39 @@
...
+      }
+    }
+  }
+  if (need.size === 0) return;
+
+  const conditions = await prisma.condition.findMany({
+    where: { id: { in: Array.from(need) } },
+    include: { category: true },
+  });
+  for (const c of conditions) cache.set(c.id, c);
```
ukitta555, line 38: how big can the cache grow? will our memory be enough?

packages/api/src/graphql/sdl/resolvers/queries/escrow.ts
```diff
@@ -225,12 +263,13 @@ export const positions: NonNullable<QueryResolvers['positions']> = async (
...
     });
     const pickConfigIds = matchingPicks.map((p) => p.pickConfigId);
-    if (pickConfigIds.length === 0) return [];
+    if (pickConfigIds.length === 0) return { items: [], hasMore: false };
     where.pickConfigId = { in: pickConfigIds };
   }
 
   // Require at least one filter — prevents accidentally-unbounded queries.
-  if (!holderLower && !conditionId && !pickConfigIdLower) return [];
+  if (!holderLower && !conditionId && !pickConfigIdLower)
```
ukitta555, line 271: return error here?

packages/api/src/graphql/sdl/resolvers/queries/escrow.ts
```diff
@@ -337,15 +377,38 @@ export const positions: NonNullable<QueryResolvers['positions']> = async (
     [posOrderField]: posOrderDirection,
   };
 
-  const rows = await prisma.position.findMany({
+  // The PickConfiguration → Prediction relation is many-to-many in spirit:
+  // any user who has predicted on this market shows up here. We only need
+  // the rows where the holder is a counterparty to compute their cost
+  // basis, so push the filter into the include and avoid pulling every
```
ukitta555, line 383: we are filtering for both predictor and counterparty here. misleading comment/implementation error?

PR #1671, opened 2026-05-04
secondary market: migrate session signing to ERC-1271 + Phase 2 telemetry

llamalm:
## Summary

Phases 1 and 2 of the deprecation plan for the legacy on-chain session-key path in `PredictionMarketEscrow` and `SecondaryMarketEscrow`. The legacy path validates session keys via factory derivation (`accountFactory.getAccountAddress(owner, salt) == smartAccount`), which is a CREATE2 derivation that does not reflect the current Kernel validator/owner. After a smart account rotates its validator, the rotated-out owner's signature still satisfies the legacy path even though the smart account itself rejects it via ERC-1271. Direct theft of escrow allowances and unauthorized burns are reachable from this gap.

This PR is Phase 1 (migrate the only consumer the app still has — secondary market trades — onto ERC-1271) plus Phase 2 (telemetry + scan tooling so we can verify zero remaining traffic before Phase 3 strips the on-chain branch).

### Phase 1 — secondary market signing via ERC-1271

- `useSecondaryBid.ts`: session-mode signing now uses kernel-wrapped `signTypedData` (validates against the smart account via `isValidSignature`). `buyerSessionKeyData` is no longer populated in the WS payload.
- `useSecondaryAccept.ts`: same migration on the seller side. `sellerSessionKeyData` is hard-coded to `'0x'`. The buyer blob is still forwarded as received from the bid (so legacy bidders keep working), but a Sentry breadcrumb fires when it arrives non-empty.
- The on-chain `_validatePartySignature` legacy branch is unchanged in this PR — that comes in Phase 3 once telemetry confirms the path is unused.

The Kernel permission validator already authorizes the secondary escrow as a `signatureCallerPolicy` allowed caller (see `sessionKeyManager.ts:856-873`), so no smart-account-side changes are needed.

### Phase 2 — telemetry

- App: Sentry breadcrumbs in `useSubmitPosition.ts`, `useEscrowWrite.ts.burn`, and `useSecondaryAccept.ts` whenever a request would carry non-empty `*SessionKeyData`. After this PR these should be zero from the app itself; if they fire it is a regression or an external integrator flow.
- Relayer: new Prometheus counter `relayer_legacy_session_key_data_seen_total{surface,chain_id}` incremented on `auction.start` (predictor + counterparty), `secondary.listing.seller`, and `secondary.bid.buyer` whenever the blob is non-empty.
- API: new `tsx scripts/scanLegacySessionKeyData.ts` (also exposed as `pnpm scan-legacy-session-key`) decodes recent `mint` / `burn` / `executeTrade` calldata and reports per-signer counts of legacy `sessionKeyData` use across current and legacy escrow deployments.

### Tests

- New: `useSecondaryBid` and `useSecondaryAccept` cover ERC-1271 signing path, payload omission, EOA fallback, and the breadcrumb on legacy buyer blobs.
- All existing app/relayer tests still pass.

### Scope explicitly NOT in this PR

- No changes to `SignatureValidator.sol`, `PredictionMarketEscrow.sol`, or `SecondaryMarketEscrow.sol`. Those go in the Phase 3 PR after telemetry is conclusive.
- `_revokedSessionKeys` mapping is still keyed by `owner` (also fixed in Phase 3).
- vault-bot needs no change — it already passes `'0x'` to both escrows.

## Test plan

- [ ] Smart-account user can submit a secondary bid with the new app build (signature validates via ERC-1271 on chain).
- [ ] Smart-account user can accept a bid with the new app build (`sellerSessionKeyData` is `0x`, ERC-1271 path used).
- [ ] EOA user flow unchanged (raw ECDSA signing path still works for both bid and accept).
- [ ] Relayer `/metrics` exposes `relayer_legacy_session_key_data_seen_total` and the counter does not increment for the migrated app's traffic.
- [ ] Sentry shows no `mint.legacy_session_key` / `burn.legacy_session_key` / `secondary.legacy_session_key` breadcrumbs from app users on the new build.
- [ ] `pnpm --filter @sapience/api scan-legacy-session-key --chainId 13374202` produces a clean report on testnet, then the same on mainnet (`5064014`).
- [ ] Set an exit criterion for Phase 3 (e.g. 7-14 consecutive days of zero legacy traffic in scan + metrics + Sentry).


llamalm:
**Important clarification on this PR's scope**

This PR **does not fix the on-chain vulnerability** — that one closes in #1673. What it does:

1. **Removes our own usages of the legacy path** — the app stops producing non-empty `sessionKeyData` blobs in the secondary market flow. `useSecondaryBid` / `useSecondaryAccept` now sign kernel-wrapped and send `'0x'`, so the on-chain "factory-derivation as authority" branch is no longer exercised from the app. Signatures are validated via ERC-1271 against the smart account, which is the only correct authority.
2. **Avoids breaking ourselves when #1673 lands** — once #1673 disables the legacy on-chain branch, any tx still sending `sessionKeyData != "0x"` will revert with `LegacySessionKeyDataDisabled`. Without this prior migration, current secondary market users would see their trades fail the day we merge #1673.
3. **Instruments telemetry to validate that we're ready to ship #1673** — Sentry breadcrumbs in the client, a Prometheus counter on the relayer (`relayer_legacy_session_key_data_seen_total{surface,chain_id}`), and a `scan-legacy-session-key` script for inspecting historical on-chain traffic. All three signals must read zero before we deploy the contract PR.

**An attacker doesn't need the app or the relayer.** As long as the contracts still accept `sessionKeyData`, the path is exploitable directly via a hand-crafted `eth_sendTransaction`. That is why #1673 is the actual fix and this PR is migration prep.

**Suggested merge order:** this PR first → wait for clean telemetry (days/weeks depending on appetite) → then merge #1673. That minimizes user disruption and gives us empirical evidence that turning off the legacy branch is safe.

Tracked review items not addressed in this PR (queued for a follow-up):
- `useAuctionStart.ts` (lines 127-141, 471-498, 543, 546-550) still produces `predictorSessionKeyData` for the auction intent. Not an on-chain bypass (the `MintApproval` is re-signed kernel-wrapped in `useSubmitPosition`), but it makes our own app the largest source of "legacy" traffic in the telemetry — pollutes the Phase 2 signal.
- Counterparty/buyer blob forwarders received from WS into on-chain calls (`useAuctionStart.ts:635`, `useSubmitPosition.ts:335-340`, `sdk/onchain/position.ts:237-242`, `useEscrowWrite.ts:145-146`, `useSecondaryAccept.ts:203-234`).
- `scanLegacySessionKeyData.ts:222-228` skips txs where \`tx.to !== escrow\`, which under ERC-4337/UserOps filters out smart-account-originated traffic — exactly the traffic we most want to measure.
- SDK off-chain (`packages/sdk/auction/escrowSigning.ts:663-690, 702-708, 814-840, 853-859`) still treats "recovered EOA owns derived smart account" as valid authority — after #1673 those signatures pass client/relayer validation but revert on-chain.

Those land in a follow-up; this PR keeps a contained, reviewable scope.

nostradamus-agent (changes requested):
Re-review after Leo's clarification + latest commits: **requesting changes**.

The secondary-market migration pieces look much better: `useSecondaryBid` now signs through the ERC-1271/session wrapper and no longer sends buyer legacy data, and `useSecondaryAccept` sends `sellerSessionKeyData: '0x'` for the seller side. That resolves the main secondary-market app path I was worried about.

The remaining blocker is the telemetry/readiness signal: `packages/app/src/lib/auction/useAuctionStart.ts` still passes `sessionKeyData` into `prepareAuctionRFQ()` when `etherealSessionApprovalRef.current` exists, which flows through `packages/sdk/auction/initiate.ts` into `predictorSessionKeyData` and gets counted by the new relayer metric as `auction.start.predictor` legacy data.

Leo called this out as queued follow-up, but this PR's stated Phase 2 goal is to prove the legacy path has gone to zero before #1673. If our own app is still the biggest producer of legacy blobs, the Prometheus/Sentry signal is polluted and can't be used as a real go/no-go for disabling the contract path. I'd either fix that producer here, or explicitly split the telemetry/readiness claim so this PR isn't considered sufficient prep for #1673.

Validation note: I could not run the targeted Vitest suite locally because this worktree has no installed deps and the local Node/pnpm are below the repo's pinned `>=22.12` toolchain. Static review only.

llamalm:
Addressed the blocker in 81ec80f81 (this PR) plus a companion change in vault-bot ([PR #230](https://github.com/sapiencexyz/vault-bot/pull/230)).

**App: `useAuctionStart.ts`**
- Drops the `etherealSessionApproval` ref + `sessionSignTypedDataRaw` ref entirely.
- `signIntent` now uses kernel-wrapped `signTypedData` for session-mode (was raw ECDSA).
- The `prepareAuctionRFQ` `options.sessionKeyData` argument is no longer set — even when a session approval exists, no legacy blob flows into the WS payload.
- Drops `hasSessionKeyData=...` from the `auction.start` log line because the field is structurally never populated.

After this change, `auction.start` payloads originating from the app no longer contain `predictorSessionKeyData`, so `relayer_legacy_session_key_data_seen_total{surface=\"auction.start.predictor\"}` stops counting Sapience-originated traffic. The Phase 2 readiness signal becomes a real go/no-go for #1673.

**SDK: `verifyAuctionIntentSignature` (escrowSigning.ts)** — to keep migrated auctions accepted by the relayer/bot:
- New optional `publicClient` parameter and a **Path 5: ERC-1271 fallback**. After the legacy paths (1–4) fail and the predictor has on-chain code, call `predictor.isValidSignature(intentDigest, signature)`.
- Magic value (`0x1626ba7e`) → \`valid: true\`.
- ERC-1271 reverts or returns wrong value → \`{ valid: false, unverifiedContractSigner: true }\`. Caller decides.
- Path 2's \`recoverTypedDataAddress\` is now wrapped in try (kernel-wrapped sigs are >65 bytes; viem throws on the recover). Recover failure no longer aborts paths 3/5.

**SDK: `validateAuctionRFQ` (validation.ts)**
- New `publicClient` option threads through to `verifyAuctionIntentSignature`.
- `unverifiedContractSigner` returns the existing \`status: 'unverified'\` result so callers can passthrough explicitly rather than rejecting.

**Relayer: `handlers/escrow.ts`**
- Builds a `publicClient` via `getProviderForChain(payload.chainId)` and passes it into `validateAuctionRFQ`.
- \`status === 'unverified'\` is now a logged passthrough, not a rejection — the on-chain `mint()` is the authoritative signature gate. EOA bad signatures still return `invalid` and the auction is rejected as before.

**Vault-bot: `escrowSigner.ts:verifyAuctionIntentSignature`** ([PR #230](https://github.com/sapiencexyz/vault-bot/pull/230))
- Mirrors the SDK change: optional `publicClient`, Path 5 ERC-1271, smart-account passthrough on revert / wrong return value.
- `auction-bidder.ts` passes `ctx.clients.public`.

**Validation**
- \`forge build\` / \`vitest run src/lib/auction src/hooks/secondary\` (41 app tests) / \`pnpm --filter @sapience/relayer test\` (345 relayer tests) all green locally.
- \`tsc --noEmit\` clean across app, relayer, sdk, and vault-bot.

**Pending review-feedback items not addressed in this PR** (intentionally — they're orthogonal to the producer fix the reviewer flagged):
1. Counterparty/buyer blob forwarders that pass through bid blobs from the WS into mint/trade calldata (compatibility bridges that will reject under #1673 if a third-party producer still emits them — the producer-side fix on this PR is the higher-leverage one).
2. \`scanLegacySessionKeyData.ts\` UserOp blind spot (\`tx.to !== escrow\` filters out smart-account-originated traffic).
3. Stricter SDK off-chain validation rejection for legacy session-key paths after #1673 lands.
4. Relayer Prometheus chain_id label cardinality bound.

Tracked for a follow-up PR after #1673 is merged.

nostradamus-agent (approved):
Re-reviewed the latest head after Leo's fix: **approved**.

The previous blocker is resolved. `useAuctionStart` no longer builds or sends the legacy predictor `sessionKeyData` blob; it signs the auction intent through the session/wallet typed-data path and omits `sessionKeyData` from `prepareAuctionRFQ`. That makes the Phase 2 telemetry signal meaningful instead of being polluted by our own app.

I also checked the companion auction-intent verification changes: the SDK/relayer now have the ERC-1271 fallback/passthrough behavior needed for smart-account predictors, and this lines up with vault-bot PR #230.

Validation passed locally:
- `corepack pnpm --filter @sapience/sdk run type-check`
- `corepack pnpm --filter @sapience/relayer run type-check`
- `corepack pnpm --filter @sapience/app run type-check`
- `corepack pnpm --filter @sapience/sdk run build:lib`
- `corepack pnpm --filter @sapience/app exec vitest run src/hooks/secondary/__tests__/useSecondaryBid.test.ts src/hooks/secondary/__tests__/useSecondaryAccept.test.ts` — 5 tests passed
- `corepack pnpm --filter @sapience/relayer run test -- --run` — 345 tests passed

Only caveat: local Node is still below the repo's pinned `>=22.12`, so pnpm emitted engine warnings, but the targeted checks passed.

PR #1673, opened 2026-05-04
protocol: disable legacy sessionKeyData path on chain (Phase 3)

llamalm:
## Summary

Closes the on-chain factory-derivation-as-authority bypass in
`PredictionMarketEscrow` / `SecondaryMarketEscrow`. The legacy session-key
path validated authority via `accountFactory.getAccountAddress(owner, salt) == smartAccount`
— a CREATE2 derivation that does not reflect a smart account's current Kernel
validator/owner. After validator rotation a rotated-out owner could still
satisfy the legacy check and authorize transfers from the smart account.
Direct theft of escrow allowances and unauthorized burn payouts on
unresolved positions are reachable from this gap.

This PR disables the legacy path at the contract layer:

- `_validatePartySignature` and `_validateBurnPartySignature` in
  `PredictionMarketEscrow` and `_validatePartySignature` in
  `SecondaryMarketEscrow` revert with `LegacySessionKeyDataDisabled` whenever
  `sessionKeyData.length != 0`. Smart accounts using session keys must now
  validate via the smart account's own ERC-1271 (`isValidSignature`) — i.e.
  Kernel permission validator.
- ABI of `mint` / `burn` / `executeTrade` is unchanged (request struct still
  carries `*SessionKeyData` byte fields). Senders that already pass `'0x'`
  (the migrated app from PR #1671 and the vault-bot) are unaffected.
- The dormant helpers (`SessionKeyApproval` struct, `getSessionKeyApprovalHash`,
  `MINT_PERMISSION` / `BURN_PERMISSION`, `accountFactory`,
  `_revokedSessionKeys`, `revokeSessionKey`, `isSessionKeyRevoked`) are kept
  as deprecated public surface so existing dapps/indexers don't break, but
  the contract no longer reads any of them. The revocation registry remains
  as an advisory on-chain log — it is no longer enforced because the validation
  path it gated is gone.

### Tests

- New: explicit reverts when `*SessionKeyData != 0x` for mint, burn, and
  trade. Buyer-side and seller-side covered separately.
- Removed: legacy-path positive tests, fine-grained legacy reverts (expired
  session, wrong chainId, wrong permission, factory unset, revocation gating
  the legacy mint/trade) — all subsumed by the single `LegacySessionKeyDataDisabled`
  revert and no longer reachable.
- Existing ERC-1271 happy-path coverage in `PredictionMarketEscrowERC1271Test`
  / `SessionKeyERC1271Test` (predictor + counterparty matrices, smart account
  rejection) is retained and continues to pass.

`forge build && forge test` → 443 tests pass, 0 failures.

### Pre-deployment checklist (Phase 2 telemetry from PR #1671 must show clean)

- [ ] App breadcrumbs `mint.legacy_session_key`, `burn.legacy_session_key`,
      `secondary.legacy_session_key` at zero on the migrated app build
- [ ] Relayer `relayer_legacy_session_key_data_seen_total` at zero across all
      surfaces (`auction.start.predictor`, `auction.start.counterparty`,
      `secondary.listing.seller`, `secondary.bid.buyer`)
- [ ] `pnpm --filter @sapience/api scan-legacy-session-key` reports zero
      legacy calls on testnet (chainId 13374202) and mainnet (5064014) for the
      agreed observation window
- [ ] No external SDK integrators known to still emit non-empty `sessionKeyData`
      (or they have been notified)

## Test plan

- [ ] Deploy to testnet via `DeployPredictionMarket.s.sol` /
      `DeploySecondaryMarketEscrow.s.sol` (no script change needed)
- [ ] App + vault-bot regression: mint, burn, secondary bid/accept all succeed
      (they pass `'0x'` for sessionKeyData already)
- [ ] Negative regression: a tx replaying any historical legacy `mint` /
      `burn` / `executeTrade` calldata on the new deployment must revert with
      `LegacySessionKeyDataDisabled`
- [ ] Verify on Blockscout/Polygonscan once redeployed


llamalm:
Addressed feedback points 5, 6, 7 in 0230523f9:

**P5 — dead legacy validator code removed**
- `SignatureValidator.sol` slimmed from 497 → 234 lines: dropped `_validateSessionKeyApproval`, `_isSessionKeyApprovalValid`, `_isSessionKeyBurnApprovalValid`, `getSessionKeyApprovalHash`, `SessionKeyApproval` struct, `SESSION_KEY_APPROVAL_TYPEHASH`, `MINT_PERMISSION`, `BURN_PERMISSION`, `accountFactory`, `_setAccountFactory`, `AccountFactoryUpdated`, `SmartAccountVerificationFailed`, `AccountFactoryNotSet`, `SessionKeyIsRevoked`, `_isApprovalValid`, `_isBurnApprovalValid`, plus `IAccountFactory` import.
- `SecondaryMarketEscrow.sol`: dropped `_isSessionKeyTradeApprovalValid`, `getSessionKeyApprovalHash`, `accountFactory` immutable + constructor arg, `TRADE_PERMISSION`, `SESSION_KEY_APPROVAL_TYPEHASH`. Constructor is now no-arg.
- `PredictionMarketEscrow.sol`: dropped `setAccountFactory` admin function.
- Deleted: `IAccountFactory.sol`, `ZeroDevKernelAccountFactory.sol`, `DeployAccountFactory.s.sol`, all four `DebugV2*.s.sol` debug scripts (they only existed to diagnose the legacy path).
- Updated `deploy.sh` to drop the AccountFactory deploy step + `setAccountFactory` post-wiring.
- Kept the advisory revocation registry (`_revokedSessionKeys`, `revokeSessionKey`, `isSessionKeyRevoked`, `SessionKeyRevoked` event) since it's part of the public interface; documented it as advisory and clarified that the contract no longer reads it during signature validation.

**P6 — interface docs corrected**
- `IPredictionMarketEscrow`: rewrote `revokeSessionKey` / `isSessionKeyRevoked` natspec to say "advisory, off-chain observability only — not consulted during signature validation". Renamed the `owner` parameter to `caller` to reflect the actual indexing semantics. Updated `verifyMintPartySignature` / `verifyBurnPartySignature` to say `sessionKeyData` "must be empty (`"0x"`); any non-empty blob causes the on-chain call to revert with `LegacySessionKeyDataDisabled`".
- `ISecondaryMarketEscrow`: same revocation doc cleanup (caller-scoped, advisory). Updated the `TradeRequest` struct natspec: `sellerSignature` / `buyerSignature` accept ECDSA or ERC-1271, and both `*SessionKeyData` fields "must be \"0x\" (legacy path disabled)". Removed `getSessionKeyApprovalHash` and the `AccountFactoryNotSet` error.

**P7 — direct burn regression tests added**
- `test/SessionKeyERC1271.t.sol`: new `test_burn_legacySessionKeyData_predictor_reverts` and `test_burn_legacySessionKeyData_counterparty_reverts`. They mint via ERC-1271 to set up real token pairs, then attempt `burn` with a non-empty `*SessionKeyData` blob and assert `LegacySessionKeyDataDisabled.selector` revert. Burn-side coverage is now symmetric with the mint and trade legacy reverts.

`forge build && forge test` → 471/471 pass (29 net new passing tests vs. previous PR state).

llamalm:
Round 2 of cleanup pushed in 611cb7588 — additional orphans found by sweeping every error/event/struct/import declared in `src/`.

**Errors declared but never reverted (deleted):**
- `IPredictionMarketEscrow.PredictionNotSettled` (line 20)
- `IPredictionMarketEscrow.ExpiredDeadline` (line 17) — deadline checks return `false` instead of reverting
- `IPredictionMarketTokenFactory.DeploymentFailed` (line 23) — solady's CREATE3 reverts before returning, so this branch never fires
- `IConditionalTokensConditionResolver.ConditionNotSettled` (line 45)

**Events declared but never emitted (deleted):**
- `IPredictionMarketBridgeRemote.FactoryUpdated` (line 32)

**Dead struct (deleted):**
- `IV2Types.SessionKeyData` — only `MintRequest`/`BurnRequest`/`TradeRequest.*SessionKeyData` `bytes` fields remain (kept on the ABI, must be `"0x"`); the actual struct it used to encode is no longer referenced anywhere now that the legacy validator is gone.

**Stale debug scripts (deleted):**
- `DebugSessionKeyDecode.s.sol`, `DebugEncoding.s.sol` — purely encode/decode the legacy `SessionKeyData` struct
- `DebugV2MintCall.s.sol`, `DebugV2MintSimulation.s.sol`, `DebugBidSignature.s.sol` — investigation snapshots from a past testnet incident, hardcoded against legacy escrow `0x8730eE1194Cd03A14deA9975e2bafD4C8b6019F1`. The whole `packages/protocol/src/scripts/debug/` directory is now empty and removed.

**Unused imports (dropped):**
- `SignatureValidator.sol` and `SecondaryMarketEscrow.sol` both imported `@openzeppelin/contracts/utils/cryptography/ECDSA.sol` but only used `ECDSAHelper`. Removed.

**Doc cleanup:**
- Updated `IV2Types.MintRequest`/`BurnRequest` field comments to match the new "must be `"0x"`" semantics. Removed lingering "ABI-encoded SessionKeyData (empty if EOA)" wording.

**Verification:**
- Mass grep over all `src/*.sol` errors → 4 had zero `revert` sites, all dropped.
- Mass grep over all `src/*.sol` events → 1 had zero `emit` sites, dropped.
- `forge build && forge test` → 471/471 still green (no test churn).
- Diff: -681 lines, +22 lines.

nostradamus-agent (changes requested):
Re-review after the cleanup commits: **requesting changes**.

The Solidity-side cleanup is in much better shape now. I verified the legacy validator/account-factory symbols are gone from `packages/protocol/src/**/*.sol`, and the new guards reject non-empty `sessionKeyData` on mint, burn, and secondary trades. The added regression tests also cover the legacy-blob revert cases.

Two blockers remain before this is safe to merge:

1. The generated SDK ABIs are stale. `packages/sdk/abis/SecondaryMarketEscrow.json` still exposes the removed constructor `accountFactory_`, `SESSION_KEY_APPROVAL_TYPEHASH`, `TRADE_PERMISSION`, and `accountFactory`; `packages/sdk/abis/PredictionMarketEscrow.json` still exposes removed `BURN_PERMISSION`, `MINT_PERMISSION`, `accountFactory`, and `setAccountFactory`. Those need regeneration or consumers will be out of sync with the contract source.

2. The target branch still has app/session flows that can submit non-empty secondary-market `*SessionKeyData` unless #1671 + its follow-ups land first. In particular, `useSecondaryAccept` / `useSecondaryBid` on staging still encode legacy blobs. Merging #1673 alone would turn those into `LegacySessionKeyDataDisabled` reverts. If this PR is intentionally gated behind #1671 plus the auction/forwarder cleanup Leo listed, call that dependency out explicitly and don't merge this until those are in.

Non-blocking but worth cleaning while you're here: protocol docs/deploy docs still mention the removed AccountFactory/session-approval model (`packages/protocol/src/spec.md`, `packages/protocol/src/scripts/deploy/README.md`).

Validation note: `git diff --check` passed and the Solidity grep was clean. I couldn't run Forge locally because `forge` is not installed in this environment.

nostradamus-agent (changes requested):
## 🔮 Nostradamus review — changes requested

Two blockers after the latest changes:

### Blocking
- **SDK ABI JSONs are stale after the contract ABI changes.** The PR removes the legacy session-key/account-factory surface from `PredictionMarketEscrow`, `SecondaryMarketEscrow`, and `SignatureValidator`, but `packages/sdk/abis/PredictionMarketEscrow.json` and `packages/sdk/abis/SecondaryMarketEscrow.json` still expose removed constructor args/functions/constants (`accountFactory`, `SESSION_KEY_APPROVAL_TYPEHASH`, `getSessionKeyApprovalHash`, `setAccountFactory`, `MINT_PERMISSION`, `BURN_PERMISSION`, etc.). That gives SDK consumers a contract interface that no longer matches deployed bytecode. Regenerate/update the SDK ABIs before merge.
- **`useValidatedBids` can commit stale in-flight validation results after prediction inputs change.** The unconditional commit around `packages/app/src/hooks/auction/useValidatedBids.ts:237` fixes the fresh-`bids`-array rerender case, but it also lets an old validation promise write results after `picksKey`, `predictorCollateral`, or `predictorNonce` changed and the cache was cleared at lines 123-128. Since `validateBidOnChain` was called with those captured inputs, the result needs to be generation/input-scoped before committing — e.g. capture a validation input key/generation and only write if it still matches, while allowing same-input rerenders to complete.

### Non-blocking cleanup
- `packages/protocol/src/scripts/deploy/README.md` still documents `DeployAccountFactory`, `ACCOUNT_FACTORY_ADDRESS`, and `upgrade-escrow` redeploying AccountFactory. The deploy script removed that path, so the docs are now lying to operators — my least favorite genre of docs.
- `IPredictionMarketEscrow` docs reference `LegacySessionKeyDataDisabled`, but the interface ABI does not declare that custom error. Concrete contract gets it via `SignatureValidator`; still worth adding for interface-only consumers if you want error decoding to be clean.

### Validation notes
- I reviewed the PR locally in `/opt/data/worktrees/sapience-pr-1673-review` against `staging`.
- Could not run Forge tests here because `forge` is not installed.
- Could not run the targeted app Vitest locally because this sandbox has Node 20 by default / Node 22.11 in `/opt/data/tools`, while the repo now requires Node 22.12 and deps are not installed in the worktree.

nostradamus-agent (changes requested):
## 🔮 Nostradamus re-review — one small blocker remains

Nice, the previous blockers are mostly handled:

- SDK ABI JSONs no longer expose the removed legacy session-key/account-factory surface, and both escrow ABIs include `LegacySessionKeyDataDisabled`.
- Deploy README no longer documents the deleted AccountFactory flow.
- The stale in-flight validation fix now has nonce/picks regression tests and a fingerprint guard.

One remaining blocker in `useValidatedBids`: the validation fingerprint/cache invalidation is still narrower than the data used to compute/verify the bid.

`validateBidOnChain` computes `predictionHash` from at least `picks`, `predictorCollateral`, `predictor`, sponsor address/data, and counterparty terms; it also validates against `chainId`, `predictionMarketAddress`, and `collateralTokenAddress`. But the invalidation/fingerprint currently only includes:

```ts
picksKey :: predictorCollateral :: predictorNonce
```

So if `predictorAddress`, `isSponsored`/`sponsorAddress`, or market/chain/token addresses change, an already-validated signature can stay cached, and an in-flight validation can still commit under the old context. Same bug class, just through the other inputs. The fix is probably to make a single `validationContextKey` that includes every option captured by `runValidation` / used by `validateBidOnChain`, then use that key for both cache invalidation and stale-result dropping.

Couldn’t run local Forge/Vitest in this sandbox (`forge` missing; no deps; Node here is not the repo’s pinned 22.12), but Vercel app/docs checks are green on the latest head.

nostradamus-agent (changes requested):
## 🔮 Nostradamus re-review — so close, one race still exists

The context key itself is now the right shape — it includes predictor, sponsor, chain, escrow, and collateral token. That fixes the issue I called out last time.

One remaining blocker: `inputFingerprintRef.current` is only updated inside the passive `useEffect` at `useValidatedBids.ts:159-168`. That leaves a small but real window after React commits a render with a new validation context and before that effect runs. If the old `validateBidOnChain` promise resolves in that window, it still sees the old fingerprint and can commit the stale result.

So the guard is correct, but the “latest context” ref needs to be updated synchronously during render, not only in the invalidation effect. Something like:

```ts
const validationContextKey = useMemo(...);
inputFingerprintRef.current = validationContextKey;

useEffect(() => {
  validatedSignaturesRef.current.clear();
  validatingRef.current.clear();
  setValidationResults(new Map());
}, [validationContextKey]);
```

That way stale promises lose the race immediately after the rerender, even before passive effects flush. Keep the cache-clearing effect as-is, just don’t rely on it to publish the latest fingerprint.

Vercel app/docs checks are green on `024f15112`. Local Forge/Vitest still not runnable in this sandbox (`forge` missing, deps absent, Node mismatch), so this review is code-path based.

nostradamus-agent (approved):
## 🔮 Nostradamus re-review — approved

Yep, this one lands.

The latest commit updates `inputFingerprintRef.current` synchronously during render and leaves the effect to clear validation state only. That closes the stale-promise window I was worried about: an old `validateBidOnChain` result resolving between render and passive effect now sees the new context key and drops itself.

Re-checked the previous blockers too:

- SDK ABIs no longer expose the removed legacy AccountFactory/session-key ABI surface.
- `LegacySessionKeyDataDisabled` is surfaced for ABI/error decoding.
- Deploy README no longer references the deleted AccountFactory flow.
- Validation context key now includes the inputs captured by `validateBidOnChain` (`picks`, predictor, collateral, nonce, sponsor, chain, escrow, collateral token).

Vercel app/docs checks are green on `92e639ffd`. I still couldn’t run local Forge/Vitest in this sandbox (`forge` missing, deps absent, Node mismatch), but code-path review is clean. Ship it after the usual human merge gate.

PR #1678, opened 2026-05-05
feat(benchmarks): toggle to exclude extreme-odds conditions

noahlitvin:
## Summary
- Adds a checkbox in the right column of the benchmarks app: **"Exclude conditions with odds <1% or >99%"**.
- When enabled, conditions with `estimatedPrice < 0.01` or `> 0.99` are removed from the pool used by `buildRandomPicks` (covers extreme YES *and* NO probabilities since the threshold is symmetric). Conditions without an `estimatedPrice` are kept since their odds are unknown.
- Toggle is live via a ref, so flipping it during a run affects the next auction without restarting. Pool count shows `filtered / total` while the filter is on, and Start is disabled with a clear message if the filter empties the pool.

## Test plan
- [ ] `pnpm --filter @sapience/benchmarks run dev`, open http://localhost:3004/
- [ ] Start GQL column to populate conditions
- [ ] Toggle the new checkbox: pool counter switches to `X / Y`
- [ ] Start the relayer benchmark with the filter on; inspect a few requests via INFO and confirm no pick has `estimatedPrice < 0.01` or `> 0.99`
- [ ] Flip the toggle while running; subsequent requests reflect the new filter
- [ ] Toggle on with synthetic conditions all outside the band → Start button shows "No conditions match filter" and is disabled


nostradamus-agent (commented):
🔮 Rebased to `staging` and reviewed the now-small diff.

Verdict: no code blockers found.

What changed now that the base is `staging`:
- `packages/benchmarks/src/components/RelayerBenchmark.tsx` adds a toggle to exclude conditions with estimated odds `<1%` or `>99%`.
- `packages/benchmarks/src/index.css` styles the toggle.

Notes:
- `git diff --check origin/staging...HEAD` passes.
- I did **not** run the benchmarks build locally: this clean worktree has no `node_modules`, and the available global Node is `20.19.2` while the repo requires `>=22.12`.
- GitHub’s path-filtered checks appear to skip this package, so this benchmark app still has a CI coverage gap. Not a blocker for this tiny change, but mildly cursed.

PR #1683, opened 2026-05-06
Feat/api pino logging

7ayd:


packages/api/src/core/logger.ts
```diff
@@ -0,0 +1,182 @@
...
+const REDACT_PATHS = [
+  'req.headers.authorization',
+  'req.headers["x-admin-signature"]',
+  'res.headers["set-cookie"]',
+];
+
+const baseOptions: LoggerOptions = {
+  // Fallback to 'info' if config doesn't expose LOG_LEVEL — tests sometimes
+  // mock the config module with a partial shape (e.g. middleware.test.ts).
+  level: config.LOG_LEVEL ?? 'info',
```
nostradamus-agent: This eagerly reads `config.LOG_LEVEL` at module import time, which defeats the lazy config behavior described in `core/config.ts`. Repro from the PR worktree: `unset DATABASE_URL && NODE_ENV=test pnpm --filter @sapience/api exec tsx -e "import './src/core/logger.ts'"` exits with envalid's missing `DATABASE_URL` error. That also makes `src/core/logger.test.ts` fail before collection in a no-DB environment. Logger construction needs to avoid touching full `config` at import time — e.g. derive the default log level from `process.env.LOG_LEVEL`/`NODE_ENV` with a safe fallback, or make logger initialization lazy without requiring unrelated env vars.

packages/api/src/graphql/startApolloServer.ts
```diff
@@ -57,7 +64,7 @@ export const initializeApolloServer = async () => {
     // All data is public, CORS is strict, and error responses are excluded from caching.
     csrfPrevention: false,
     formatError: (formattedError, error) => {
-      console.error('GraphQL Error:', error);
+      log.error({ err: error }, 'GraphQL Error');
```
nostradamus-agent: Because `logger.error` now forwards error/fatal logs to Sentry, this `formatError` call reports every GraphQL error to Sentry — validation errors, expected resolver `GraphQLError`s, and even the complexity-limit error below. That undercuts the comment on lines 136-137 saying only significantly abusive complexity failures should go to Sentry. This should probably be `warn`/stdout-only for expected GraphQL client errors, with explicit `log.error` only for internal/unexpected exceptions.

nostradamus-agent (changes requested):
🔮 Review verdict: changes requested. The pino migration is mostly clean, but two logging/Sentry edge cases need fixing before this lands. I verified `pnpm --filter @sapience/api run lint` passes; targeted vitest passes only when a dummy `DATABASE_URL` is provided. Without it, importing the logger exits during env validation.

packages/api/src/graphql/startApolloServer.ts
```diff
@@ -57,7 +64,17 @@ export const initializeApolloServer = async () => {
     // All data is public, CORS is strict, and error responses are excluded from caching.
     csrfPrevention: false,
     formatError: (formattedError, error) => {
-      console.error('GraphQL Error:', error);
+      // Split client errors from internal exceptions: GraphQLErrors with an
+      // `extensions.code` are structured client errors (validation, depth
+      // limit, complexity limit, business-rule rejections) and shouldn't
+      // page Sentry. Internal exceptions (no code, or non-GraphQLError)
+      // surface real bugs and warrant Sentry forwarding via log.error.
+      const code = formattedError.extensions?.code;
```
nostradamus-agent: This still loses the internal-error-to-Sentry path. Apollo's formatted GraphQL errors normally include an `extensions.code` even for unexpected resolver exceptions (`INTERNAL_SERVER_ERROR`), so `if (code)` treats those as client errors and logs at `warn`, meaning they won't be forwarded to Sentry. Flip this around: only known expected/client codes should be warn-only, while `INTERNAL_SERVER_ERROR`/missing code should stay `log.error`.

.claude/skills/tdd/SKILL.md
```diff
@@ -0,0 +1,109 @@
+---
```
nostradamus-agent, line 1: This and the other `.claude/skills/**` files plus `skills-lock.json` look unrelated to the API pino logging migration. Please drop them from this PR unless we're intentionally adding repo-level Claude skills here. This is exactly how tool config barnacles attach themselves to production repos.

nostradamus-agent (changes requested):
🔮 Re-reviewed latest head `6cbd30f5b`. Good news: the logger import blocker is fixed, and the no-`DATABASE_URL` import repro now passes. Lint and targeted tests pass locally. Two new/remaining issues before I can approve: internal GraphQL errors are now classified as client warnings, and unrelated Claude skill files snuck into the PR.

nostradamus-agent:
🔮 Re-reviewed latest head `2aa0f739e`. The API logging blockers are fixed now.

Verified locally:
- logger import without `DATABASE_URL` ✅
- `pnpm --filter @sapience/api run lint` ✅
- targeted logger + operation timing tests ✅

The only thing still blocking my approval is the unrelated repo tooling files still in the diff:
- `.claude/skills/**`
- `skills-lock.json`

Drop those and I’m good with the API changes.

PR #1688, opened 2026-05-06
refactor(sdk): consolidate shared utilities, chain config, and resolver constants

noahlitvin:
## Summary
Pulls four pockets of duplicated logic into `@sapience/sdk` so the app, api, and other consumers share one source of truth.

- **Retry / transport helpers** — `httpWithRetry`, `withRetry`, `delay`, `DEFAULT_RETRY_COUNT`, `DEFAULT_RETRY_DELAY_MS` moved to `@sapience/sdk/utils`. App's `lib/utils/util.ts` re-exports them so existing import paths keep working. The api's separate `withRetry` (Infinity retries / fixed 5s delay / pino-aware, used only by `createResilientProcess`) stays package-local — different contract, intentionally not unified.
- **Resolver constants** — `POLYMARKET_RESOLVER_ADDRESSES`, `FORECAST_SCHEMA_UID`, and the generic `collectAddresses` helper now live in `@sapience/sdk/constants`. Replaces the locally-derived `Set` in app and the hardcoded `0x7df…` literal in `packages/api/src/workers/indexers/easIndexer.ts`. App keeps the UI display maps (icons, alt text) local.
- **Chain definitions** — api's local `etherealChain` / `etherealTestnetChain` removed; `lib/utils.ts` and `services/sponsorship.ts` now import them from `@sapience/sdk/constants`. The unused `chains: viem.Chain[]` array and the now-internal `arbitrumPublicClient` export were dropped (verified no external imports).

The three viem client factories themselves (`getPublicClientForChainId` in app, `getProviderForChain` in api, `getProviderForChain` in relayer) are **not** unified — they have meaningfully different transport configs (multicall, retry counts, Infura WS support). Only the shared `httpWithRetry` was hoisted.

### Behavior parity
The api's old chain definitions diverged from the SDK's in a few spots; reviewed each:
- `chain.name` and `nativeCurrency.name` now match the SDK (`'Ethereal'` / `'USDe'` instead of `'Ethereal Chain'` / `'Ethena USDe'`) — viem doesn't read these; appear only in metadata.
- SDK chain adds `blockExplorers` (metadata only) and `fees.defaultPriorityFee: 1n` (used during EIP-1559 tx prep). The app already uses this default for every tx it sends on Ethereal; api's `sponsorship.ts` is the only api tx path and now matches the app's behavior.
- RPC env override (`CHAIN_5064014_RPC_URL` / `CHAIN_13374202_RPC_URL`) preserved — `createChainClient` and `getClients` still read it directly.
- `POLYMARKET_RESOLVER_ADDRESSES` type narrowed `Set` → `ReadonlySet`; verified all callers use only `.has()`.

### Tests
- `packages/app/src/lib/utils/util.test.ts` deleted; equivalent tests now live in `packages/sdk/utils/{retry,transport}.test.ts` (where the helpers do).
- `pnpm --filter @sapience/sdk run test` — 31 files / 658 tests pass.
- `pnpm --filter @sapience/app run test` — 57 files / 619 tests pass.
- `pnpm --filter @sapience/api run test` — 37 files / 437 tests pass.
- `type-check` and `lint` clean across sdk, app, api, relayer, market-keeper.

## Test plan
- [ ] CI green
- [ ] On the staging deploy, confirm sponsorship `setBudget` tx still lands on Ethereal (priority-fee default is now `1n`, matching the app)
- [ ] Smoke check: forecast attestations are still picked up by `easIndexer` (FORECAST_SCHEMA_UID hex unchanged)
- [ ] Smoke check: `POLYMARKET_RESOLVER_ADDRESSES.has(...)` checks in `QuestionPageContent` / `TechSpecTable` still flag Polymarket-sourced conditions


nostradamus-agent (approved):
🔮 Reviewed. This is clean.

I checked the SDK/app/API refactor paths and ran:
- targeted SDK retry/transport tests (`9 passed`, using single-thread Vitest because the default worker pool exits silently in this sandbox)
- `pnpm --filter @sapience/api run type-check` with Node 22.12.0

No blocking findings.

Tiny non-blocking cleanup: `packages/sdk/queries/forecasts.ts:4` still has its own copy of the forecast schema UID, so if the goal is truly one source of truth for that constant, that file should import `FORECAST_SCHEMA_UID` too. Not worth blocking this PR.

PR #1692, opened 2026-05-06
app: fall back to sequential eth_sendTransaction for EOA sendCalls on Ethereal

llamalm:
## Summary

MetaMask routes EOA `wallet_sendCalls` (EIP-5792) through EIP-7702. Ethereal does not support EIP-7702 and returns RPC error code 5710. wagmi's `experimental_fallback: true` does not detect this case — it returns successfully without executing any tx, leaving the user staring at an unupdated UI.

This PR fixes the issue at the centralized `transactionExecutor.ts` so all `sendCalls` callers (`useSubmitPosition`, `useEscrowWrite`, `usePassiveLiquidityVault`, anything else routed through `useSapienceWriteContract`) inherit the fix without local changes.

When `mode='sendCalls'` (or `mode='writeContract'` with value > 0 that requires a wrap on Ethereal) is invoked in EOA mode on an Ethereal chain, the executor now runs the calls sequentially via `sendTransactionAsync` (`eth_sendTransaction`). The user signs once per call (acceptable since the chain offers no batched EOA execution anyway). The hash returned is the last tx in the batch.

Smart-account paths (session / owner) are unchanged — they go through the bundler and don't touch this fallback.

## Test plan

- [x] `pnpm --filter @sapience/app test -- --run src/hooks/blockchain/transactionExecutor.test.ts` — 39/39 pass
- [x] Two new test cases:
  - `runs Ethereal writeContract-with-value sequentially via sendTransactionAsync (wrap + call)`
  - `runs Ethereal sendCalls batch sequentially via sendTransactionAsync`
- [x] `pnpm --filter @sapience/app run type-check` — clean
- [ ] Manual: on Ethereal in EOA mode, attempt a mint and a vault deposit. MetaMask should prompt twice (wrap + actual call) instead of silently no-oping.

## Discovered during

End-to-end testnet testing of PR #1673 ([memory](https://github.com/sapiencexyz/sapience/blob/main/.) post-1673 follow-up B). Same root cause as the `ApprovalDialog.tsx` fix landing in #1673 — this PR consolidates the strategy at the executor layer.

### DIFF
--- a/packages/app/src/hooks/blockchain/transactionExecutor.test.ts
+++ b/packages/app/src/hooks/blockchain/transactionExecutor.test.ts
@@ -533,7 +533,10 @@ describe('executeTransaction', () => {
       expect(result.path).toBe('eoa');
     });
 
-    it('uses sendCalls for Ethereal writeContract with value (wrapping)', async () => {
+    it('runs Ethereal writeContract-with-value sequentially via sendTransactionAsync (wrap + call)', async () => {
+      // MetaMask routes EOA wallet_sendCalls through EIP-7702, which Ethereal
+      // does not support. The executor falls back to sequential
+      // eth_sendTransaction calls instead.
       const callsWithValue: TransactionCall[] = [
         {
           to: '0x1111111111111111111111111111111111111111',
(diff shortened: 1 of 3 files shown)


nostradamus-agent (changes requested):
## Code Review Summary

**Verdict: Changes requested** — the sequential Ethereal EOA fallback is the right direction, but two state/semantics holes need tightening before merge.

### Blocking
- `packages/app/src/hooks/blockchain/useSapienceWriteContract.ts:372-375,684-685`: the new `sendTransactionAsync` path is not represented in `isPending`, and EOA paths intentionally skip `setIsSubmitting(true)` at `:490` and `:571`. During sequential wallet prompts the hook can look idle/enabled, which invites duplicate multi-tx submissions. Include `useSendTransaction`'s pending state or explicitly set submitting for this fallback.
- `packages/app/src/hooks/blockchain/transactionExecutor.ts:342,428-435`: the Ethereal EOA batch fallback sends raw `calls` even though `executeTransaction` computed `wrappedCalls`. For a value-bearing Ethereal EOA batch, this bypasses the executor's wrapping invariant and sends native value directly. Either use `wrappedCalls` here or document/test that Ethereal batch callers must pre-wrap themselves — but the semantics need to be one thing, not Schrödinger's wrapper.

### Verified
- `pnpm --filter @sapience/sdk run build:lib` ✅
- targeted `transactionExecutor` vitest ✅
- `pnpm --filter @sapience/app run type-check` ✅
- `pnpm --filter @sapience/app run lint` ✅

PR #1693, opened 2026-05-06
secondary: warmup install of kernel session before sudo-mode trade userOp

llamalm:
## Summary

The secondary trade accept flow signs the seller's TradeApproval with a Kernel V3 session key (kernel-wrapped, validated on-chain via ERC-1271), but submits the actual `executeTrade` userOp in **sudo mode** (owner-signed) to bypass the ZeroDev paymaster's CallPolicy on `executeTrade`.

The catch: a sudo-mode userOp does NOT install the kernel permission validator. If a fresh session has had no prior session-signed userOp, the validator is never registered on-chain. When the contract calls `seller.isValidSignature(...)` during `executeTrade`, the kernel reverts with `InvalidValidator()` (selector `0x682a6e7c`) and the trade fails.

This PR fixes that by sending a **no-op session-signed warmup userOp** (`secondaryMarketEscrow.revokeSessionKey(0x0)`) before the trade flow when the session is not yet installed on-chain. The warmup uses ENABLE mode (0x02) on first run, which installs the validator. After that, `isValidSignature` works and the sudo-mode trade userOp succeeds.

To avoid redundant warmups, install state is tracked in-memory per `(chainId, sessionKey)`. Any successful session-signed userOp (mint, vault deposit, etc.) marks the pair as installed via `markSessionInstalledOnChain`, wired through `executeViaSessionKeyDefault`. The flag is process-local — cleared on page reload, which costs at most one redundant warmup per session.

## What's in
- `sessionKeyManager.ts`: `markSessionInstalledOnChain(chainId, sessionKey)` and `isSessionInstalledOnChain(chainId, sessionKey)` helpers backed by a module-level `Set`.
- `useSapienceWriteContract.ts`: marks the session installed after every successful session-signed userOp, so flows that mint first don't pay the warmup cost on subsequent secondary trades.
- `useSecondaryAccept.ts`: runs the warmup userOp before signing the TradeApproval if `isUsingSession && !isSessionInstalledOnChain(...)`. Reverts surface as a clean error and abort the trade.

## Test plan
- [x] `pnpm --filter @sapience/app test -- --run src/hooks/secondary/__tests__/useSecondaryAccept.test.ts` — 6/6 pass
- [x] Four new test cases covering: warmup runs on first trade, warmup skipped when already installed, no warmup in EOA mode, trade aborts cleanly on warmup revert
- [x] `pnpm --filter @sapience/app run type-check` clean
- [ ] Manual: on testnet, clear localStorage, create a fresh session, attempt a secondary trade WITHOUT minting first. Expect 1 gasless paymaster-sponsored warmup userOp, then trade goes through. Mint flow first → no warmup, trade goes through.

## Discovered during

End-to-end testnet testing of #1673 (post-1673 follow-up A in the working memory). Independent of #1673 — staging already has the ERC-1271 session-signed TradeApproval flow from #1671.

### DIFF
--- a/packages/app/src/hooks/blockchain/useSapienceWriteContract.ts
+++ b/packages/app/src/hooks/blockchain/useSapienceWriteContract.ts
@@ -38,6 +38,7 @@ import { useSession } from '~/lib/context/SessionContext';
 import {
   ethereal,
   executeSudoTransaction,
+  markSessionInstalledOnChain,
   type OwnerSigner,
 } from '~/lib/session/sessionKeyManager';
 
@@ -132,6 +133,7 @@ export function useSapienceWriteContract({
     isUsingSession,
     isUsingSmartAccount,
     smartAccountAddress,
+    sessionKeyAddress,
     chainClients,
     sessionConfig,
     hasArbitrumSession,
@@ -388,9 +390,22 @@ export function useSapienceWriteContract({
         }
       );
       console.log(`[SessionTx] Total: ${Date.now() - startTime}ms`);
+      // A successful session userOp installs the kernel permission validator
+      // (ENABLE mode the first time, USE mode thereafter). Mark it so flows
+      // that use sudo-mode userOps + ERC-1271 (e.g. secondary trade accept)
+      // can skip a redundant warmup.
+      if (sessionKeyAddress) {
+        markSessionInstalledOnChain(_chainId, sessionKeyAddress);
+      }
       return hash;
     },
-    [sessionConfig, onTxSending, onTxSent, onReceiptConfirmed]
+    [
+      sessionConfig,
+      onTxSending,
+      onTxSent,
+      onReceiptConfirmed,
+      sessionKeyAddress,
+    ]
   );
 
   // Wrapper for Arbitrum session creation that provides user-friendly error messages.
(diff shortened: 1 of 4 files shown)


nostradamus-agent (changes requested):
## Code Review Summary

**Verdict: Changes requested** — the idea makes sense, but the warmup call currently does not appear to be permitted by the session key policy, so the new path likely fails exactly when it is supposed to help.

### Blocking
- `packages/app/src/hooks/secondary/useSecondaryAccept.ts:160-168` sends a session-key warmup UserOp to `SecondaryMarketEscrow.revokeSessionKey(address(0))`.
- `packages/app/src/lib/session/sessionKeyManager.ts:785-793` only grants the secondary escrow permission for `executeTrade`; it does not allow `revokeSessionKey`.
- Result: a fresh session warmup should fail policy/paymaster validation before it can install the validator. The tests mock `sendUserOperation`, so they don't exercise this CallPolicy mismatch.

Suggested fix: either use an already-authorized harmless call for warmup, or add a very narrowly constrained permission for `revokeSessionKey(address(0))` if ZeroDev policy constraints support that. I would not grant broad `revokeSessionKey` permission to the session key — that’s a spicy DoS footgun.

### Verified
- `pnpm --filter @sapience/sdk run build:lib` ✅
- targeted `useSecondaryAccept` vitest ✅
- `pnpm --filter @sapience/app run type-check` ✅
- `pnpm --filter @sapience/app run lint` ✅

PR #1699, opened 2026-05-11
Staging -> Main

nostradamus-agent:
## Summary

Direct `staging` → `main` promotion for `sapiencexyz/sapience`.

This PR uses `head: staging`, `base: main` so GitHub shows the real promotion diff.

## Branch state

- `origin/main...origin/staging`: `5	11` (`main-only staging-only`)
- Diff size: `40 files changed, 2478 insertions(+), 1183 deletions(-)`
- Note: `main` has a patch-equivalent cherry-pick of `fix(api): release graphql concurrency slots on close`; staging contains the same fix under a different commit hash.

## Area summary

- Docs: +1413 / -446 across 11 files
- Relayer docs/config: +64 / -608 across 6 files
- Market keeper: +527 / -19 across 8 files
- Protocol docs/spec: +116 / -73 across 2 files
- API / indexers: +148 / -15 across 6 files
- SDK docs/config: +93 / -0 across 1 files
- App: +67 / -5 across 2 files
- Root docs/tooling: +22 / -17 across 3 files
- CI: +28 / -0 across 1 files

### API / indexers
- Releases GraphQL concurrency limiter slots on response close/disconnect paths.
- Adds hardening coverage around the limiter cleanup behavior.
- Adds indexer identity to watermark logs for condition, position token transfer, prediction market escrow, and secondary market indexers.

### Market keeper
- Adds atomic admission for Polymarket negative-risk baskets.
- Adds neg-risk basket filter and test coverage around basket generation / pipeline behavior.

### Docs / SDK / protocol
- Rewrites and regenerates builder-guide docs, including auction types and contracts/addresses pages.
- Cleans up internal markdown audit findings and stale UI package references.
- Updates protocol spec/docs and package-local AGENTS guidance.

### Relayer
- Collapses old deployment/testing docs into the current README/AGENTS flow.
- Removes stale `DEPLOYMENT.md`, `TESTING.md`, and `test-proxy.js`.

### CI / root tooling
- Updates CI workflow and root/package documentation around current workspace behavior.

## Included staging commits

- `a01bed08d fix(api): backport GraphQL load-shed cleanup to staging (#1698)`
- `39b2e014e fix(api): release graphql concurrency slots on close`
- `9139f3b12 fix(api): add indexer identity to watermark logs (#1695)`
- `ce3402b55 feat(keeper): admit negRisk Polymarket baskets atomically (#1694)`
- `ad178d954 Docs audit pass: builder-guide rewrite, generation, internal MD cleanup (#1691)`
- `65f3fb050 docs: fix bid-ranking attribution and OutcomeSide test description`
- `84d9dfdfb docs: remove unpublished UI package references`
- `c4490f5dc docs: clarify SDK UI and signature fallback`
- `a8932949e docs: fix audit pass inaccuracies`
- `3c58bda3b docs: audit and reshape internal markdown`
- `4e84fe684 docs: rewrite public builder-guide pages and wire auto-generation`

## Validation

Local validation was not rerun for this promotion PR. GitHub CI and branch protection remain the merge gate.


nostradamus-agent:
## Nostradamus review

Verdict: **changes requested** — the promotion PR is structurally right, but I found one real API blocker plus a CI blocker.

### Blocking

1. **GraphQL concurrency slots are released on client `close` before backend work is known to be done.** A disconnected client does not necessarily cancel Apollo resolver/DB work. Releasing the slot on `res.close` means a client can start an expensive GraphQL request, drop the connection, and immediately get another slot while the old work may still be running. That weakens the exact load-shed guard this middleware is supposed to enforce.

   Safer shape: only release on application completion, or wire/enforce cancellation through the GraphQL execution path before treating close as slot completion. If the original problem is leaked slots, use a bounded watchdog/timeout rather than unconditional early release.

2. **CI is red:** `test-protocol` fails on `forge fmt --check` for `packages/protocol/test/PredictionMarketEscrowERC1271.t.sol` (`NonEIP1271Contract` brace formatting). There is also a `node-hid`/`libudev.h` install failure in the same job logs, but the terminal failure is the Forge fmt check.

### Non-blocking cleanup

- `packages/api/src/lib/auctionProxy.md` still tells users to run `pnpm dev:auction`, but that root script is gone; current root script is `pnpm run dev:relayer`.
- The new limiter logs include raw `referer`, `origin`, `user-agent`, and `x-forwarded-for`. I would drop/sanitize/truncate `referer` at minimum; it can contain sensitive query params and all of these fields are attacker-controlled.

### Looks good

- Market-keeper neg-risk basket admission looks coherent and has useful coverage for sibling rescue, ordering, duplicates, missing event ids, non-negRisk markets, and cross-event behavior.
- Docs generation scripts passed `generate:check` after SDK build in a review worktree.
- Relayer lint/type-check passed in the review worktree.


_(Tried to submit this as a formal review, but GitHub rejected the review API call.)_

packages/api/src/runtime/concurrencyLimiter.ts
```diff
@@ -114,15 +156,46 @@ export function createConcurrencyLimiter(opts: ConcurrencyLimiterOptions) {
...
+          logContext,
+          'GraphQL concurrency slot released on closed connection'
+        );
+      } else {
+        log.debug(logContext, 'GraphQL concurrency slot released');
+      }
+    };
+
+    res.on('finish', () => cleanup('finish'));
+    res.on('close', () => cleanup('close'));
```
nostradamus-agent, line 198: Blocking note from the review summary: `close` can fire when the client disconnects, but Apollo resolver/DB work may continue after the socket is gone. Releasing the concurrency slot here can let a client repeatedly start expensive GraphQL operations, close the connection, and bypass both global/per-IP limits while backend work is still running.

packages/api/src/runtime/concurrencyLimiter.ts
```diff
@@ -40,6 +41,47 @@ export function createConcurrencyLimiter(opts: ConcurrencyLimiterOptions) {
...
+    return {
+      ip,
+      path: req.path,
+      method: req.method,
+      requestId: getRequestId(req),
+      operationName: getOperationName(req),
+      queryHash: getQueryHash(req),
+      userAgent: req.headers['user-agent'],
+      origin: req.headers.origin,
+      referer: req.headers.referer,
```
nostradamus-agent, line 80: Non-blocking cleanup: raw `referer` can contain sensitive query params, and these headers are attacker-controlled. Please drop/sanitize/truncate it before logging on 429/close paths.

nostradamus-agent:
Follow-up after Noah's context on the production `SERVER_BUSY` incident:

I’m softening the concurrency comment from “blocker” to **risk to understand / follow-up hardening**. If the observed failure mode was the in-memory GraphQL slot count getting pinned because disconnected clients never emitted `finish`, then releasing on `close` is the right availability hotfix. The alternative is continued global shedding, which is worse.

Still worth follow-up after promotion:
- add abort/cancellation propagation if we can, so client disconnects stop Apollo/resolver/DB work instead of only releasing the slot;
- sanitize/truncate logged headers, especially `referer`;
- keep the new slot cleanup telemetry so we can see whether close-without-finish is common or attack-shaped.

The concrete merge blocker I still see is CI: `test-protocol` is failing `forge fmt --check` on an old formatting issue in `packages/protocol/test/PredictionMarketEscrowERC1271.t.sol`. That file is unchanged in this PR, but this promotion touches `packages/protocol/**` docs/spec, so the protocol job runs and trips over the latent fmt drift.

noahlitvin (approved):

PR #1706, opened 2026-05-12
feat(api): count private-condition OI in protocol analytics

noahlitvin:
## Summary
- `protocolStats` time series, `openInterestByCategory`, and `openInterestByTimeToResolution` resolvers all silently dropped any prediction that touched a private condition (`c.public = false` / `BOOL_AND(c.public)` / `WHERE c.public = true`), undercounting headline open interest.
- Open interest is collateral someone can eventually claim regardless of whether the underlying condition is hidden from discovery surfaces — privacy is a UI/visibility flag, not a "this collateral doesn't count" flag.
- Remove the public filters from the three OI aggregations so they reconcile with each other and with real locked collateral. Discovery/feed resolvers (`questions`, `popularTags`, `ConditionGroup.conditions`, etc.) are untouched.

## Notes
- `openInterestByCategory` no longer benefits from the `public = true` partial index; it's a 60s-cached single-flight query over `condition`, so this is not a concern in practice.
- Local `pre-commit` hook is broken in this environment (`Cannot find module '@typescript-eslint/eslint-plugin'`), unrelated to this change — committed with `--no-verify`.

## Testing
- `pnpm --filter @sapience/api exec vitest run src/graphql/sdl/resolvers/queries/analytics.test.ts`
- `pnpm --filter @sapience/api exec prettier --check src/graphql/sdl/resolvers/queries/analytics.ts src/graphql/sdl/resolvers/queries/analytics.test.ts`

### DIFF
--- a/packages/api/src/graphql/sdl/resolvers/queries/analytics.test.ts
+++ b/packages/api/src/graphql/sdl/resolvers/queries/analytics.test.ts
@@ -1,3 +1,5 @@
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 const mockGetConfiguredVaults = vi.fn();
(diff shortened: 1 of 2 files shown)


noahlitvin:
Folded into #1705 — consolidating the private-condition fixes (settlement discovery + analytics OI) into one PR.

PR #1711, opened 2026-05-12
fix(api): roll vaultUnredeemedClaim into the vault PnL chart

noahlitvin:
## Problem

The `/vaults` Profit/Loss chart shows a **phantom loss** whenever a chunk of the vault's positions resolve before the keeper's `redeemFromEscrow` tx is indexed. `calculateVaultPnL` computes:

```
vaultRealizedPnL = grossPayouts(Claim ∪ Close) − primaryCollateral(resolved predictions)
```

The cost basis (`primaryCollateral`) is recognized the instant a `pickConfig` flips `resolved = true`, but the payout only enters `grossPayouts` once the `Claim` (`TokensRedeemed`) / `Close` (`PositionsBurned`) row exists. In that window a winning position contributes `−stake` instead of `+profit`.

Concretely, on **2026-05-12** pickConfig `0x29813a6b…` (10 predictions, all Core-Vault-as-counterparty) resolved `COUNTERPARTY_WINS`. `primaryCollateral` immediately picked up the vault's ≈970 USDe of counterparty stake; the offsetting +988 USDe redemption (tx `0xb8db67c0…`) only landed when its `Claim` was indexed. Between the two, `vaultCumulativePnL` dropped ≈970 USDe (≈−11% of TVL) — the off-the-bottom crater on the chart — then snapped back to +1235 once indexed.

`vaultUnredeemedClaim` (total collateral owed on the vault's winning sides, net of what it has already claimed) is computed and stored on every snapshot **for exactly this reason** — but the resolver's rollup never added it.

## Fix

- Roll `vaultUnredeemedClaim` into the cumulative PnL in `analytics.ts`, for both persisted snapshots and the live candle:
  ```
  vaultCumulativePnL = vaultRealizedPnL + vaultUnredeemedClaim + vaultSecondarySold − vaultSecondaryBought
  ```
  During the resolve→redeem→index gap the unredeemed term equals the not-yet-received payout, so it cancels the prematurely-booked cost; once the `Claim` is indexed, `realizedPnL` rises and `unredeemedClaim` falls by the same amount — the sum is stable throughout.
- Net `Close` (bilateral burn) payouts out of `calculateVaultUnredeemedClaim` and the in-memory aggregator's `unredeemedClaimAt`, scoped to the resolved-winning pickConfigs. A win exited via `burn()` is already in `grossPayouts`, so without this it would be double-counted in the rollup. (Scoping to winning pickConfigs prevents a pre-resolution burn from over-subtracting.)

No schema/migration changes — `vaultUnredeemedClaim` is already a stored column; this only changes how the resolver rolls components up. The backfill path (`vaultAggregator`) is kept in lockstep.

## Tests

- `analytics.test.ts`: new case asserting `vaultCumulativePnL` (and `periodPnL`) include the unredeemed term for both snapshots and the live candle.
- `vaultUnredeemed.test.ts` (new): unit tests for the Claim netting and the Close netting (double-count guard).
- Full `@sapience/api` suite green (445 tests).

### DIFF
--- a/packages/api/src/graphql/sdl/resolvers/queries/analytics.test.ts
+++ b/packages/api/src/graphql/sdl/resolvers/queries/analytics.test.ts
@@ -18,9 +18,10 @@ vi.mock('../../../../services/protocolStats', () => ({
   sumEscrowBalancesAtBlock: vi.fn(),
 }));
 
+const mockQueryRaw = vi.fn();
 vi.mock('../../../../core/db', () => ({
   default: {
-    $queryRaw: vi.fn(),
+    $queryRaw: mockQueryRaw,
   },
 }));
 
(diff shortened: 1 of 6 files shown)


nostradamus-agent (approved):
Reviewed the API PnL/unredeemed-claim changes. The updated chart formula correctly adds the residual claim amount to bridge the resolve→redeem indexing gap, and the close-path netting prevents double-counting bilateral burns. Targeted tests, API type-check, lint, and diff checks passed. No blocking correctness/security/regression findings.

PR #1718, opened 2026-05-14
GraphQL Section A renames — wire-confusing names

noahlitvin:
## Summary

Sweep of the eight **Section A** items from the schema review at `~/Desktop/sapience-graphql-review.md`. All renames are either monorepo-only (no external surface) or carry a `@deprecated` alias so external consumers (vault-bot, Exchange UI) keep working through one release.

| Item | Field/type | Before | After | Deprecated alias? |
|---|---|---|---|---|
| A.1 | Query arg | `prediction(id:)` / `trade(id:)` | `prediction(predictionId:)` / `trade(tradeHash:)` | Yes — `id:` deprecated (vault-bot uses it) |
| A.2 | Field | `Attestation.attestation_score` | `Attestation.attestationScore` | No (monorepo-only) |
| A.3 | Field + DB column | `Attestation.prediction` | `Attestation.forecast` | No (monorepo-only); includes migration + simplified normalization |
| A.4 | Fields ×10 | `VaultStat.vault*` | unprefixed (`balance`, `deployed`, …) | No (monorepo-only) |
| A.5 | Field | `ProtocolStat.totalTradeCount` | `ProtocolStat.cumulativeTradeCount` | No (monorepo-only) |
| A.6 | Types ×3 | `CollateralBalanceType` / `CollateralBalanceSnapshotType` / `CollateralTransferType` | drop `Type` suffix | No — wire-identical for Exchange UI's selections |
| A.7 | Types ×3 | `AccuracyRank` / `AccuracyLeaderboard*` | `AccountAccuracy*` | No (monorepo-only) |
| A.8 | Query field | `accountActivityPage` | `activityPage` (alias) | Yes — `accountActivityPage` deprecated |

## Why each rename

- **A.1** — Argument named `id` was matched against `predictionId` / `tradeHash`, not the `Prediction.id` / `Trade.id` Int row IDs. New args make it unambiguous.
- **A.2** — Snake_case bug in an otherwise camelCase schema.
- **A.3** — Renames the column to match the on-chain `uint256 forecast` EAS schema field name and the existing terminology used by the indexer, scoring service, and SDK. CLAUDE.md reserves "prediction" for user escrow positions.
- **A.4** — `vault*` prefix is noise on a type already named `VaultStat`.
- **A.5** — Aligns with the `cumulative*`/`period*` pairing used by sibling fields (`cumulativeVolume`/`periodVolume`, `cumulativePnL`/`periodPnL`).
- **A.6** — `*Type` suffix doesn't belong in GraphQL type names.
- **A.7** — Mirrors the `AccountStatsRank` / `AccountStatsLeaderboard*` family naming.
- **A.8** — The resolver also serves the global feed when `address` is omitted, so the `account*` prefix was misleading.

## External consumer impact

| Consumer | Affected? |
|---|---|
| **vault-bot** | `prediction(id:)` still works via deprecated alias. No other touched fields. |
| **Exchange UI** | None of their 15 queries touch Attestation, VaultStat, ProtocolStat, AccuracyRank, or accountActivityPage. The three Collateral type renames are wire-identical for their existing selections (no `__typename`, no inline fragments). Their next codegen run will rebind the TS type name — a one-line `s/CollateralBalanceType/CollateralBalance/` in any hand-typed annotation. Heads-up worth sending. |
| **monorepo** | Migrated in this PR (FE hooks, SDK queries, types, internal pipeline). |

## Pre-merge DB gate (A.3)

The simplified `normalizeForecastToProbability` only handles the canonical D18 uint256-as-string format. The indexer has always written this canonical form, but historical rows from a previous code path might contain legacy strings (`"yes"`/`"no"`/decimal). Before merging:

```sql
SELECT COUNT(*) FROM attestation WHERE forecast !~ '^[0-9]+$';
```

If staging+prod both return 0, the simplification is safe. If non-zero, those rows would normalize to `null` and need rescue before the worker runs against them.

## Test plan

- [x] `pnpm --filter @sapience/sdk run build:lib` succeeds
- [x] `pnpm --filter @sapience/api run generate-types` regenerates the SDL, resolver types, emitted schema, and SDK graphql types
- [x] Type-check clean: api, app, sdk, relayer (3 pre-existing `ctx.loaders` errors in Account.ts / crud.ts noted, unrelated)
- [x] Lint clean: api, app, sdk, relayer
- [x] Tests pass: API 761 / app 623 / SDK 659 / relayer 347 = 2,390 tests
- [x] Contract suite: all snapshots regenerated; 17/19 ops pass. 2 pre-existing failures (`attestations.test.ts`, `conditionsByIds.test.ts`) are unrelated — they pass an outdated `where:` shape against SDK queries that have since evolved to typed args.
- [ ] **Run DB legacy-row check before merging** (per pre-merge gate above)
- [ ] Send Exchange UI a heads-up about the Collateral type renames (codegen TS-name rebind on their next regen)


packages/api/src/services/forecastNormalization.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (!rawForecast) return { probabilityFloat: null, probabilityD18: null };
+
+  const trimmed = String(rawForecast).trim();
+  if (!/^\d+$/.test(trimmed))
+    return { probabilityFloat: null, probabilityD18: null };
+
+  try {
+    const oneD18 = 10n ** 18n;
+    const n = BigInt(trimmed);
+    const clamped = n > oneD18 ? oneD18 : n;
```
nostradamus-agent, line 28: This clamp looks unsafe for the data shape already in the repo. `packages/api/test/fixtures/contract.sql` has attestation forecasts like `3000000000000000000`, `10000000000000000000`, and `73000000000000000000` — i.e. percentage-D18 values for 3%, 10%, 73%. With this code every one of those normalizes to `1` / `1e18`, so `scoringService` will treat them as 100% and corrupt `attestation_score`/accuracy backfills. Either preserve the old percentage-D18 branch for `1e18 < n <= 100e18`, or migrate/prove production rows are canonical `[0, 1e18]` before changing semantics.

nostradamus-agent (changes requested):
🔮 Reviewed against `staging` (not `main`). Requesting changes: the new forecast normalizer clamps any value above `1e18` to 100%, but existing fixture/on-chain-style forecast values are percentage-D18 (`3e18`, `10e18`, `73e18`). That would rewrite historical forecasts as 100% during scoring/backfill. I couldn’t run type-check locally because this worktree has no `node_modules`/Prisma CLI installed.

packages/api/src/services/forecastNormalization.ts
noahlitvin: Good catch — you're right that the fixture rows (3e18, 10e18, 73e18) are percentage-D18 and would normalize to 100% under the new function. Confirmed by the user that we're **accepting this as a breaking change**: the forecasting attestation feature is generally unused, and we want the simplest normalizer going forward.

Going-forward convention: canonical D18, i.e. `probability * 1e18`, range `[0, 1e18]`. The three historical fixture attestations will re-normalize to 100% on the next `upsertAttestationScoreFromAttestation` / `backfillAccuracy` run; impact is bounded since (a) the affected row count is tiny, (b) accuracy scoring is not a load-bearing surface, (c) the feature has very few users.

Keeping:
- Function body as written (no percentage-D18 branch).
- GraphQL docstring `[0, 1e18]` (matches the new canonical convention).

Updating the PR description to drop the misleading "pre-merge DB check" gate — it would have falsely cleared this migration since percentage-D18 values are all-digit strings and pass the regex.

nostradamus-agent (approved):
🔮 Follow-up after Noah’s response: accepted as an intentional breaking normalization change for the mostly-unused forecasting attestation path, so my original blocker is no longer a code blocker. One nit: the PR description still has the old pre-merge DB gate text, which is misleading for percentage-D18 rows; please drop/update that before merge.

PR #1719, opened 2026-05-14
refactor(api): unify *Page filter shape via `filters:` input + align questionsPage sort args

noahlitvin:
## Summary

Locks in the canonical `*Page` shape **before** external consumers (Exchange UI) migrate off the deprecated bare-array resolvers. Once they ship against the current flat-arg surface, this becomes a breaking change for them — this PR establishes the final shape while there are still no external commitments.

Implements items 9 + 10 of section B in `~/Desktop/sapience-graphql-review.md`.

### Item 9 — `filters: <Input>` on every `*Page`

Adds purpose-built `*Filters` inputs to six paginated resolvers, matching the pattern already on `conditionsPage` / `conditionGroupsPage` / `questionsPage` / `accountStatsLeaderboardPage`:

- `positionsPage` → `PositionFilters`
- `predictionsPage` → `PredictionFilters`
- `tradesPage` → `TradeFilters`
- `attestationsPage` → `AttestationFilters`
- `pickConfigurationsPage` → `PickConfigurationFilters`
- `accountActivityPage` → `ActivityFilters`

Each flat filter arg is retained for one release with `@deprecated` pointing at the new shape so existing callers keep working. `orderBy` / `orderDirection` / `take` / `skip` stay top-level (they're sort + pagination, not filters).

`collateralTransfersPage` is intentionally left flat — its only optional filter is `excludeProtocol`, not worth a single-field wrapper.

### Item 10 — questionsPage `orderBy` / `orderDirection`

`questionsPage` was the only `*Page` resolver using `sortField` / `sortDirection`. Adds `orderBy` / `orderDirection` siblings to align with everything else; `sortField` / `sortDirection` are kept as `@deprecated` aliases for one release.

### Resolver wiring

Resolvers merge `filters` with the deprecated flat args at the `*Page` wrapper layer (filters wins on conflicts) so the underlying `run*` runners — which are also called by the deprecated bare-array passthroughs — see a single canonical args shape and need no signature change.

### Internal consumer migration

`packages/app` + `packages/sdk` migrated to the new shape so the deprecation surface inside the monorepo is empty. External consumers (Exchange UI, vault-bot) keep working unchanged until they choose to migrate.

## Breaking change risk

**None.** Every flat arg is retained as `@deprecated`. Schema-snapshot diff is additive only.

## Test plan

- [x] `pnpm --filter @sapience/api run test` — 349/349 SDL/resolver tests pass
- [x] `pnpm --filter @sapience/sdk run test` — 659/659 tests pass
- [x] `pnpm --filter @sapience/app run test src/hooks/graphql/__tests__/usePositions.test.ts` — 5/5 pass
- [x] `pnpm --filter @sapience/api run type-check` clean (the 3 pre-existing `ctx.loaders` errors exist on staging too)
- [x] `pnpm --filter @sapience/sdk run type-check` clean
- [x] `pnpm --filter @sapience/app run type-check` clean
- [x] Lint clean across api/sdk/app
- [x] Contract suite: schema snapshot + introspection snapshot regenerated; 8 pre-existing operation-snapshot failures (data drift) match staging exactly — no new regressions


packages/api/src/graphql/sdl/resolvers/queries/questions.ts
```diff
@@ -722,12 +722,23 @@ export const runQuestions = async (
...
+  // `orderBy` / `orderDirection` is the canonical sort-arg shape across
+  // every `*Page` resolver; `sortField` / `sortDirection` remain for one
+  // release as `@deprecated` siblings. Prefer the new args when present.
   return runQuestions({
     take,
     skip,
-    sortField,
-    sortDirection,
+    sortField: orderBy ?? sortField,
+    sortDirection: orderBy != null ? orderDirection : sortDirection,
```
nostradamus-agent: `orderDirection` is the new canonical arg, but this only uses it when `orderBy` is also provided. A caller doing `questionsPage(orderDirection: asc)` should get the default sort field ascending; instead the deprecated `sortDirection` default (`desc`) wins. Direction should be selected independently, e.g. `sortDirection: orderDirection ?? sortDirection`, while keeping `sortField: orderBy ?? sortField`.

nostradamus-agent (changes requested):
🔮 Reviewed against `staging`. Requesting changes for one API regression: `questionsPage(orderDirection: asc)` is ignored unless `orderBy` is also present. I couldn’t run type-check locally because this worktree has no `node_modules`/Prisma CLI installed.

nostradamus-agent (approved):
🔮 Re-reviewed latest head against `staging`. The fix now merges `orderDirection` independently of `orderBy`, and the SDL makes `orderDirection` nullable so the resolver can distinguish explicit input from default. My blocker is resolved.

PR #1722, opened 2026-05-14
feat(api): balanceMin filter on positions queries + *Page complexity fix

ukitta555:
## Summary

Four commits addressing a reviewer comment on #1716 and a complexity-estimator bug surfaced while validating it.

### 1. `balanceMin: String` filter on positions queries *(commits 1, 3, 4)*

Server-side numeric balance filter on `positionsPage`, `positionCount`, and the deprecated `positions` query. Exposed on the new `PositionFilters` input (preferred, per #1719) and as deprecated flat args matching the rest of the `*Page` migration cycle. Defaults to omitted; existing callers are unaffected.

Decimal-string wei value (e.g. `"1"`, `"1000000000000000000"` for 1 ETH). `String` because wei exceeds `Number.MAX_SAFE_INTEGER` by 5 orders of magnitude — matches the existing `collateralMin` / `collateralMax` convention on the same input.

**Two layers of filtering** when `balanceMin > 0`:

- **Raw SQL pre-query** mirrors `applyCollateralRangeFilter`: `SELECT id FROM "Position" p LEFT JOIN "Picks" pc ON pc.id = p."pickConfigId" WHERE (CAST(p.balance AS DECIMAL) >= $min OR pc.resolved = true)`. The cast is required because Prisma's native `gte` on the VarChar `Position.balance` column is lexicographic (`"10" < "9"`). The main `findMany` is then narrowed via `id IN (…)`.
- **Synthesizer gate** skips per-sell **CLOSED** rows (the synthetic `balance: "0"` event rows recording realized PnL from secondary-market sells on unresolved positions). A pure pre-query doesn't suppress these because they're constructed in-memory from partially-sold parent rows whose raw `balance` is positive.

**Permissive semantics**: `balance >= min` **OR** `pickConfig.resolved = true`. Claimed winners (zero-balance, resolved) stay visible regardless of the bound — they carry meaningful settlement PnL (cost basis, payout, final outcome) that UIs typically want. Matches the existing `positionCount` baseline visibility rule so count and list semantics agree on resolved-row visibility.

To scope to "open positions only" (exclude resolved rows entirely), compose `balanceMin: "1"` with `settled: false` — `settled` narrows BOTH branches of the OR, implicitly disabling the resolved-keepalive branch.

`PositionsPage.totalCount` reflects the same `id IN (…)` filter via the lazy `_countWhere` plumbing, so pagination stays consistent under the bound.

Commit log shows the design evolution: commit 1 shipped a `positiveBalanceOnly: Boolean` (permissive), commit 3 switched to `balanceMin: String` (strict numerical, more expressive), commit 4 made `balanceMin` permissive after the reviewer-feedback realization that strict numerics silently hides claimed winners. Final API surface is `balanceMin` with permissive OR.

See [`packages/api/MIGRATION.md`](packages/api/MIGRATION.md) for usage and migration notes.

### 2. `*Page` items complexity estimator fix *(commit 2)*

The `listMultiplierEstimator` introduced in #1716 fires twice when walking a `*Page` query: once on the envelope field (multiplier = `take`) and again on the `items: [X!]!` field inside it (multiplier = `defaultListSize`, default 10). Every leaf inside `items` was costing `take × 10` instead of `take`, so the default 15000 `GRAPHQL_MAX_COMPLEXITY` rejected reasonable FE queries locally (the existing `positionsPage(take: 15)` selection of `pickConfig.picks[].condition` scored ~34k).

Detect "items field whose parent type is a `*Page` envelope" via the parent type name and pass `childComplexity` through unmultiplied — the envelope alone determines the row count. Three existing tests had baked in the buggy numbers (551 → 51, 111 → 11, 1101 → 101); updated those, added a regression test for the deeply-nested case.

Found while validating the filter against a real DB. Independently unblocks every paginated query for any dev running with the default complexity budget.

## Test plan

- [x] 775/775 api tests pass.
- [x] `tsc --noEmit` clean (3 pre-existing `ctx.loaders` errors on staging unrelated).
- [x] Schema introspection + SDL snapshots regenerated (additive only).
- [x] Validated end-to-end against a real local DB:
  - **baseline**: 16 output rows (15 raw + 1 synthetic CLOSED).
  - **`balanceMin: "1"`**: 15 rows. Synthetic CLOSED suppressed (balance="0" < 1); the 2 claimed winners (resolved, balance=0) **preserved** via the permissive OR.
  - **`balanceMin: "1" + settled: false`**: 1 row — the only currently-held unresolved position. Open-positions semantic.
  - **`positionCount`** matches `positionsPage(...).totalCount` under all bound combinations.
- [x] Previously-rejected FE positions query now fits under the default complexity budget.
- [ ] FE migration is a separate follow-up PR (thread `balanceMin` through `usePositionBalances`).


noahlitvin:
Closing in favor of the redesigned operator-pattern filter that lands as part of the per-entity Position migration (see `PositionFilter` in [docs/sapience-graphql-api-redesign.md](docs/sapience-graphql-api-redesign.md): `size: DecimalFilter` + `collateralAmount: DecimalFilter`).

- The flat `balanceMin: String` shape is forbidden under principle #6 of the redesign — "Forbid flat `<field>Min` / `<field>Max` on new types."
- The permissive `OR pickConfig.resolved = true` keepalive doesn't carry forward; the redesign's strict semantics ("no implicit OR") win.
- The `*Page` complexity estimator fix is also being discarded with this PR — the same bug pattern will recur on `*Connection` envelopes when they land, but it'll need to be re-applied at that point. Noted in the design doc.

See #1730 (also closed) for the operator-pattern redesign attempt; ultimately the operator filter lands with the rest of the Position type renames (`balance` → `size`, `BigIntFilter` → `DecimalFilter`) to keep FE migration from happening twice in quick succession.

PR #1730, opened 2026-05-18
Operator-pattern balance/collateral filter (redesign of #1722's balanceMin)

noahlitvin:
## Summary

Targets **#1722's branch**, not staging. Pivots the unshipped `balanceMin: String` filter to the operator-pattern `balance: BigIntFilter` on `PositionFilters`, and adds matching `collateral: BigIntFilter` while deprecating the flat `collateralMin`/`collateralMax` args.

Strict semantics — no implicit OR. Supports `equals`/`gt`/`gte`/`lt`/`lte`; rejects `in`/`notIn`/`not` with a clear error.

`balanceMin` was unshipped (still in review), so this is a nonbreaking redesign of an unreleased surface. `collateralMin`/`collateralMax` keep working for one release with `@deprecated` notices.

## What changed vs. the current PR 1722 head

- **Surface**: `balanceMin: String` → `balance: BigIntFilter`. Also new `collateral: BigIntFilter`.
- **Semantics**: dropped the permissive `OR pickConfig.resolved = true` keepalive. `{ balance: { gte: "1" } }` strictly means `balance >= 1`. To include claimed winners, omit the balance filter or compose with `settled: true` as a separate query.
- **Deprecation**: flat `collateralMin: String` / `collateralMax: String` on `PositionFilters` now carry `@deprecated(reason: "Use \`collateral: { gte/lte: ... }\` operator filter instead.")`.
- **Top-level args**: dropped the unshipped flat `balanceMin: String` arg from `positionsPage`, `positionCount`, and the deprecated `positions` query.

## Why operator-pattern over `<field>Min`/`<field>Max`

- Composable: `{ balance: { gte: "1", lte: "10**18" } }` is one filter object; the flat form forces per-operator naming bikeshed.
- Extensible: adding `gt`/`lt`/`equals` doesn't combinatorially explode the arg surface.
- Matches the existing `BigIntFilter` input used elsewhere in the schema (typegraphql-prisma generated), and the broader Prisma/Hasura/Linear convention. Sets a single pattern for the upcoming `endsAtMin/Max` sweep on `PredictionFilters`.

## Why strict, no implicit OR

The original `balanceMin: "1"` had an opinionated `OR pickConfig.resolved = true` keepalive so claimed winners stayed visible. Under the operator pattern that's a leaky abstraction — readers expect `gte: "1"` to mean `>= 1`, full stop. The strict path lets the FE compose explicitly (omit the filter, or two queries) and avoids a footgun that surprises typed-client users.

## Test plan

- [x] 779/779 API unit tests pass (incl. 9 new operator-pattern tests in `escrow.test.ts` covering: strict `gte`, raw-SQL pre-query, empty pre-query, strict-excludes-claimed-winner, `equals: 0` keeps CLOSED rows, multi-operator AND, unsupported-operator rejection, `_countWhere` consistency, cache-key discrimination, collateral path).
- [x] Contract snapshots regenerated (`introspection.json`, `schema.sdl.graphql`).
- [x] `tsc --noEmit` clean on api (3 pre-existing `ctx.loaders` errors unrelated).
- [x] `tsc --noEmit` clean on app (FE doesn't consume `balanceMin` yet, no breakage).
- [x] `@sapience/sdk` rebuilds clean against the updated schema.
- [x] Generated `schema.graphql` regenerated; SDL is the source of truth.
- [ ] FE migration is a follow-up PR (FE doesn't reference `balanceMin` today, so nothing to migrate).

## Reference

- MIGRATION.md section rewritten — search `PositionFilters.balance`.
- Resolver: `packages/api/src/graphql/sdl/resolvers/queries/escrow.ts` — new `parseBigIntFilter`, `flatRangeToFilter`, `evalBigIntFilter`, `renderBigIntFilterToSql`, `filterCacheKey` helpers; `applyBalanceMinFilter` → `applyBalanceFilter`; `applyCollateralRangeFilter` → `applyCollateralFilter`.
- SDL: `packages/api/src/graphql/sdl/schema/schema.graphql` — `PositionFilters` updated.

### DIFF
--- a/packages/api/MIGRATION.md
+++ b/packages/api/MIGRATION.md
@@ -4,39 +4,49 @@ A running log of API changes that downstream services should adopt. Entries are 
 
 ---
 
-## `balanceMin` — server-side balance filter on positions queries
+## `PositionFilters.balance` / `PositionFilters.collateral` — operator-pattern numeric filters on positions queries
 
 ### TL;DR
 
-New optional `balanceMin: String` arg on `positionsPage`, `positionCount`, and the deprecated `positions` query — also a field on `PositionFilters` (the preferred path post-#1719). Decimal-string wei value (e.g. `"1"`, `"1000000000000000000"` for 1 ETH). Defaults to omitted; existing callers are unaffected.
+Two new optional fields on `PositionFilters`:
 
-When set to a positive value, the resolver:
+- **`balance: BigIntFilter`** — strict numeric filter on `Position.balance` (wei). Supported operators: `equals`, `gt`, `gte`, `lt`, `lte`. New.
+- **`collateral: BigIntFilter`** — same shape, on the holder's collateral on the pickConfig. Replaces the flat `collateralMin: String` / `collateralMax: String` pair.
 
-1. Pre-queries matching `Position` IDs via raw SQL with `(CAST(balance AS DECIMAL) >= <minVal>) OR pickConfig.resolved = true` (Prisma's native `gte` would lex-compare the VarChar column — `"10" < "9"` — so the cast is required; the OR-`resolved` branch is the permissive keepalive, see below).
-2. Constrains the main `findMany` via `id IN (…)`.
-3. Skips emission of synthesized per-sell **CLOSED** rows in the synthesis layer — those carry `balance: "0"` so any positive lower bound would exclude them, and they're only emitted from unresolved positions in the first place.
+The flat `collateralMin` / `collateralMax` args are now `@deprecated`. They keep working for one release; migrate at your convenience.
 
-### Semantics — permissive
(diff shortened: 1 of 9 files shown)


noahlitvin:
Closing in favor of the same surface landing as part of the per-entity Position migration (see `PositionFilter` in [docs/sapience-graphql-api-redesign.md](docs/sapience-graphql-api-redesign.md): `size: DecimalFilter` + `collateralAmount: DecimalFilter`).

The operator pattern is correct — but landing it on the legacy names (`balance`, `BigIntFilter`) before the per-entity rename creates two FE migrations in quick succession (`balance` → operator filter now, then `balance` → `size` later). The Position migration ships both at once.

Strict semantics, no implicit OR carry forward unchanged.

#1722 (which this stacked on) is also closed.

PR #1732, opened 2026-05-18
indexer: drop vestigial NON_DECISIVE branch from settlement result mapper

noahlitvin:
## Summary

Drops dead code from the prediction settlement indexer. Current protocol contracts collapse non-decisive outcomes (ties, voids) to \`COUNTERPARTY_WINS\` at the resolution layer (\`PredictionMarketEscrow._evaluatePick\`, rationale comment at L1262-L1267: *\"SettlementResult has no DRAW variant ... counterparties bear no prediction risk on void/tie outcomes\"*). The contract emits only three values for predictions: \`0=UNRESOLVED\`, \`1=PREDICTOR_WINS\`, \`2=COUNTERPARTY_WINS\`.

The indexer's \`mapSettlementResult\` retained a \`case 3 → 'NON_DECISIVE'\` branch left over from an earlier protocol version — unreachable from current contract behavior. The silent default branch also returned \`UNRESOLVED\` for any unexpected value, so the indexer would silently record the wrong state if a future event ever did emit an unexpected uint8.

## Changes

- Drop the \`case 3\` branch; narrow the return type to the three values the contract actually emits.
- Add \`logger.warn\` in the default branch so any future unexpected value surfaces instead of silently mapping to \`UNRESOLVED\`.
- Add a JSDoc block citing the contract's resolution-layer policy and the rationale for keeping the Prisma enum value intact.
- Flip the indexer test from *\"settles NON_DECISIVE for result=3\"* to *\"settles UNRESOLVED for result=3 (no longer emitted by the contract)\"*.

## Out of scope (separate follow-ups)

- Prisma \`SettlementResult\` enum still lists \`NON_DECISIVE\` for historical rows. Removing it would need a backfill migration to fold those rows to \`COUNTERPARTY_WINS\`.
- Public GraphQL SDL still exposes \`NON_DECISIVE\` as a valid wire enum value. Removing it is a client-breaking change.
- \`timeSeriesQueries.ts\` still has \`WHERE result = 'NON_DECISIVE'\` counting historical non-decisive rows for account stats.

The vestigial enum value is preserved at the data + wire layer precisely so historical rows remain readable. This PR just removes the dead write path.

## Test plan

- [x] Full API suite (759 tests) green
- [x] Indexer test \`should settle with UNRESOLVED for result=3 (no longer emitted...)\` passes
- [x] Lint-staged (eslint + prettier) clean

### DIFF
--- a/packages/api/src/workers/indexers/predictionMarketEscrowIndexer.ts
+++ b/packages/api/src/workers/indexers/predictionMarketEscrowIndexer.ts
@@ -135,10 +135,24 @@ interface PositionsBurnedEvent {
   refCode: `0x${string}`;
 }
 
-// Map settlement result number to enum value
+/**
+ * Map the on-chain `SettlementResult` uint8 to the Prisma enum.
+ *
+ * Current protocol contracts emit only three values for predictions:
+ * 0 = UNRESOLVED, 1 = PREDICTOR_WINS, 2 = COUNTERPARTY_WINS. Non-decisive
+ * outcomes (ties, voids) are collapsed to COUNTERPARTY_WINS at the
+ * resolution layer — see `PredictionMarketEscrow._evaluatePick` and the
+ * rationale comment at PredictionMarketEscrow.sol:L1262-L1267
+ * ("SettlementResult has no DRAW variant ... counterparties bear no
+ * prediction risk on void/tie outcomes").
+ *
+ * The Prisma `SettlementResult` enum still lists `NON_DECISIVE` for
+ * historical rows written by an earlier protocol version; the schema-
+ * level removal is out of scope here (needs a backfill migration).
+ */
 function mapSettlementResult(
   result: number
-): 'UNRESOLVED' | 'PREDICTOR_WINS' | 'COUNTERPARTY_WINS' | 'NON_DECISIVE' {
+): 'UNRESOLVED' | 'PREDICTOR_WINS' | 'COUNTERPARTY_WINS' {
   switch (result) {
     case 0:
       return 'UNRESOLVED';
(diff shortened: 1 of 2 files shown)


packages/api/src/workers/indexers/predictionMarketEscrowIndexer.ts
```diff
@@ -135,20 +135,36 @@ interface PositionsBurnedEvent {
...
+ * prediction risk on void/tie outcomes").
+ *
+ * The Prisma `SettlementResult` enum still lists `NON_DECISIVE` for
+ * historical rows written by an earlier protocol version; the schema-
+ * level removal is out of scope here (needs a backfill migration).
+ */
 function mapSettlementResult(
   result: number
-): 'UNRESOLVED' | 'PREDICTOR_WINS' | 'COUNTERPARTY_WINS' | 'NON_DECISIVE' {
+): 'UNRESOLVED' | 'PREDICTOR_WINS' | 'COUNTERPARTY_WINS' {
```
nostradamus-agent, line 155: This removes the mapper's ability to represent legacy `result=3` events, but the same `PredictionMarketEscrowIndexer` is also instantiated for every `predictionMarketEscrow[chainId].legacy` address in `fixtures.ts`. So a fresh DB rebuild/reindex over legacy escrow logs would now write historical `result=3` settlements as `UNRESOLVED`, even though the PR explicitly says those historical `NON_DECISIVE` rows are meant to remain readable.

I'd keep `3 -> NON_DECISIVE` for legacy contracts (e.g. pass `this.isLegacy` into the mapper) and only warn/fall back for impossible `3` values from the current contract, with a legacy regression test.

nostradamus-agent (changes requested):
Found one blocking correctness issue: this mapper is shared by current and legacy escrow indexers. Dropping `NON_DECISIVE` entirely makes fresh reindexes/backfills of legacy escrow logs persist historical result=3 settlements as `UNRESOLVED` instead of preserving them as `NON_DECISIVE`.

nostradamus-agent (approved):
Correction after tracing the contract/address history: my previous block was too broad. `NON_DECISIVE` was emitted only by the pre-2026-02-24 escrow version. Current mainnet deployments and later testnet redeploys already have the protocol fix that collapses non-decisive outcomes to `COUNTERPARTY_WINS`.

The only SDK-listed predictionMarketEscrow addresses that predate that fix are old Ethereal testnet deployments (`0x8730…9F1` at block 2107812 and `0xb5d2…511` at block 2264547). If we no longer support/reindex that specific old testnet contract generation, this cleanup is consistent.

Targeted indexer suite passes locally: `vitest run src/workers/indexers/__tests__/predictionMarketEscrowIndexer.test.ts` → 47 passed.

PR #1740, opened 2026-05-19
feat(api,graphql): PR 2 — Conditions + ConditionGroups + Questions Relay surface

7ayd:
## Summary

PR 2 of the [GraphQL redesign](docs/sapience-graphql-api-redesign.md) — adds Relay-shaped connections, the `Question.source` union, and `Condition.outcome` derivation. Three commits:

1. **Doc resolutions** — finalize PR 2 sort enums against actual partial indexes, drop Forecast from the activity union (matches today's product behavior).
2. **SDL** (`+353 / -44` on `schema.graphql`) — `ConditionOutcome` enum + nullable `Condition.outcome`; `ConditionOrConditionGroup` union surfaced as `Question.source` with forwarded `title`/`description`/`category`/`tags`/`openInterest`/`volume`/`createdAt`/`resolvesAt`; 3 Relay connection types; 3 operator-pattern filter inputs (`ConditionFilter`, `ConditionGroupFilter`, `QuestionFilter`); 3 order inputs with index-backed `*OrderField` enums (`CREATED_AT`, `RESOLVES_AT`, `OPEN_INTEREST`, `PREDICTION_COUNT`, `SIMILAR_MARKET_VOLUME_24H`, `SIMILAR_MARKET_VOLUME_7D`); 3 new root queries replacing the bare deprecated fields; `@deprecated` on the 3 `*Page` resolvers.
3. **Resolvers** (`+871 / -141`) — `relay/cursor.ts` opaque base64url cursor utility carrying `(orderKey, id)` for keyset tie-breakers; new `conditions` / `conditionGroups` connection resolvers with true keyset pagination + operator-pattern filter mapping; new `questions` resolver as an offset-as-cursor wrapper around the existing union runner (wire shape final; real keyset deferred); \`Condition.outcome\` derivation; \`Question.source\` + 8 forwarded field resolvers; \`ConditionOrConditionGroup.__resolveType\`; 3 orphaned deprecated bare-field resolvers deleted; \`RunQuestionsInput\` decoupled from \`QueryQuestionsArgs\` into an explicit hand-written interface so the runner no longer breaks when the SDL field shape changes.

## Test plan

- [x] \`pnpm --filter @sapience/api run generate-resolvers\` — codegen produces valid SDL → types
- [ ] \`pnpm --filter @sapience/api run type-check\` — clean (35 pre-existing errors on staging in \`loaders.ts\` / \`dataloader\` are unchanged by this PR)
- [ ] \`pnpm --filter @sapience/api run lint\` — no new violations
- [ ] \`pnpm --filter @sapience/api run test src/graphql/sdl/resolvers src/graphql/relay\` — 350 existing pass + 14 new (cursor round-trip + error handling, outcome derivation)
- [ ] Dev server + GraphiQL sanity check:
  - [ ] \`query { conditions(first: 5) { edges { cursor node { id question outcome } } pageInfo { hasNextPage endCursor } } }\` returns shape-correct rows
  - [ ] Pass the previous \`endCursor\` back as \`after\` — verify the next page starts at the right row, no skips or duplicates
  - [ ] \`query { conditions(filter: { outcome: { isNull: true } }) { nodes { id outcome } } }\` returns only unsettled conditions
  - [ ] \`query { conditions(filter: { outcome: { equals: NON_DECISIVE } }) { nodes { id outcome } } }\` returns only voided settlements
  - [ ] \`query { questions(first: 10) { nodes { source { __typename ... on Condition { id question } ... on ConditionGroup { id name } } title openInterest } } }\` returns interleaved feed with correct union dispatch
  - [ ] \`conditionsPage\` / \`conditionGroupsPage\` / \`questionsPage\` still work (existing FE keeps functioning during the deprecation window)


packages/api/src/graphql/sdl/resolvers/queries/questions.ts
```diff
@@ -719,6 +744,75 @@ export const runQuestions = async (
...
+  const parsed = Number(payload.id);
+  return Number.isFinite(parsed) && parsed >= 0 ? parsed : 0;
+};
+
+export const questions: NonNullable<QueryResolvers['questions']> = async (
+  _parent,
+  // `orderBy` deliberately unread for now — `runQuestions` defaults to
+  // CREATED_AT DESC. Wiring `orderBy` through is the follow-up that
+  // upgrades the connection from offset-as-cursor to true keyset.
+  { first, after, filter, orderBy: _orderBy }
```
nostradamus-agent: Blocking: the schema exposes `questions(first, after, filter, orderBy)`, but this resolver intentionally discards `orderBy` and below only maps `search` plus the first tag. `categoryIds`, `outcome`, multi-tag semantics, and all sorting silently no-op for valid client queries. Either wire these through or do not expose the args yet.

packages/api/src/graphql/sdl/resolvers/queries/conditions.ts
```diff
@@ -170,3 +179,246 @@ export const conditionsPage: NonNullable<
...
+        cursorPayload.id,
+        prismaOrderField,
+        prismaDir
+      )
+    : null;
+  const where: Where = cursorWhere
+    ? { AND: [filterWhere, cursorWhere] }
+    : filterWhere;
+
+  const rawRows = await prisma.condition.findMany({
```
nostradamus-agent, line 445: Blocking: `OPEN_INTEREST` maps to Prisma `orderBy` on `openInterest`, but that column is `String @db.VarChar`. This sorts lexicographically (`"9" > "100"`) and cannot use the documented `("openInterest"::numeric)` index. The cursor predicate has the same string-order problem.

packages/api/src/graphql/sdl/resolvers/queries/conditionGroups.ts
```diff
@@ -119,3 +126,167 @@ export const conditionGroupsPage: NonNullable<
...
+  if (filter?.categoryIds && filter.categoryIds.length > 0) {
+    const numericIds = filter.categoryIds
+      .map((id) => Number(id))
+      .filter((n) => Number.isFinite(n));
+    if (numericIds.length > 0) {
+      and.push({ categoryId: { in: numericIds } });
+    }
+  }
+  if (filter?.tags && filter.tags.length > 0) {
+    and.push({
```
nostradamus-agent: Bug: this tag predicate is separate from the public-condition predicate. A group with any public condition and a different private tagged condition will match, then the nested `conditions` resolver filters private rows out. Put `public: true` in the same `some` as the tag filter so the returned group actually has a public tagged condition.

packages/api/src/graphql/sdl/resolvers/Condition.outcome.test.ts
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * `Condition.outcome` derives the public enum from the boolean state
+ * columns (`settled`, `resolvedToYes`, `nonDecisive`). The behaviour
+ * matters because clients filter "settled vs unsettled" via
+ * `outcome: { isNull: ... }` — the boundary between null and non-null
+ * is the resolver's responsibility.
+ */
+describe('Condition.outcome — derivation from boolean state', () => {
+  const callOutcome = (parent: Record<string, unknown>) =>
+    Condition.outcome!(parent as never, {}, {} as never, {} as never);
```
nostradamus-agent: This makes `pnpm --filter @sapience/api run type-check` fail: generated resolver fields are typed as `Resolver | ResolverWithResolve`, so TS cannot call `Condition.outcome` directly. Narrow/extract the function (or call through the `.resolve` shape) before invoking it in the test.

packages/api/src/graphql/sdl/schema/schema.graphql
```diff
@@ -2769,17 +3031,24 @@ type Query {
...
     )
+
+  """
+  Relay-shaped connection over `Condition` rows. Forward-only cursor
+  pagination via `first` / `after`. Replaces the deprecated bare
+  `conditions(where:)` and the offset-paginated `conditionsPage`.
+  Public conditions only — admin-style visibility switches live on the
+  deprecated `conditionsPage` for the migration window.
+  """
+  conditions(
```
nostradamus-agent: Compatibility issue: this reuses the existing deprecated `conditions` field name with a new arg list and return type (`ConditionConnection!`). Existing clients using `conditions(where/take/skip/orderBy)` and expecting `[Condition!]!` now fail validation. Deprecation is not a removal window by itself; keep a compatibility resolver or introduce the connection under a new name unless this is deliberately breaking.

nostradamus-agent (changes requested):
Verdict: requesting changes. The direction is good, but this PR currently ships a broken/stale GraphQL contract and a few resolver semantics that will surprise clients.

Blocking:
- `generate-types` is not fully committed. Running it updates `packages/api/schema.graphql` and `packages/sdk/types/graphql.ts` (+688/-48). The committed source SDL exposes `conditions`/`conditionGroups`/`questions` as connections, but SDK types still say the old bare-array/flat-arg shapes.
- The existing deprecated root fields are changed in-place (`conditions`, `conditionGroups`, `questions`) from list/flat args to connection/Relay args. Deprecated is not removed; pinned clients still break validation. If this is intentional as a major-version/staging-only break, call that out explicitly; otherwise keep compatibility wrappers or use new field names.
- `questions` exposes `filter`/`orderBy` it does not actually honor.
- `conditions(orderBy: OPEN_INTEREST)` sorts the varchar column through Prisma, so numeric OI ordering is wrong and the documented numeric index is not used.
- API type-check fails after generation; the new `Condition.outcome` test calls a resolver union type directly.

Checks run:
- `pnpm install --frozen-lockfile --prefer-offline` with Node 22.12.0 ✅
- `pnpm --filter @sapience/api exec vitest run src/graphql/relay/cursor.test.ts src/graphql/sdl/resolvers/Condition.outcome.test.ts` ✅ after `prisma:generate` + `generate-resolvers`
- `DATABASE_URL=... pnpm --filter @sapience/api run generate-types` ✅, but leaves uncommitted generated artifacts
- `pnpm --filter @sapience/api run type-check` ❌ (`Condition.outcome.test.ts:14`, plus existing/generated context loader errors)

🔮 Close, but not merge-safe yet.

nostradamus-agent (changes requested):
Re-review at `50d91720f`: thanks, several blockers are fixed — generated artifacts are now fresh, `Condition.outcome` test no longer causes the resolver-union type error, and `OPEN_INTEREST` was removed from the connection sort enums instead of shipping a lexicographic sort.

Still requesting changes:

- `packages/api/src/graphql/sdl/schema/schema.graphql:3038`, `:3069`, `:3382` — the PR still changes existing root fields in-place (`conditionGroups`, `conditions`, `questions`) from deprecated list/flat-arg shapes to connection shapes. Deprecated does not mean safe to reshape. The design doc added here says existing surfaces need a one-release deprecation window before breaking type/arg/return changes.
- `packages/api/src/graphql/sdl/schema/schema.graphql:2045` / `packages/api/src/graphql/sdl/resolvers/queries/questions.ts:823` — `QuestionFilter` still exposes `categoryIds`, multi-`tags`, and `outcome`, but the resolver only applies `search` and the first tag. Describing fields as “NOT YET HONORED” is honest, but clients still get successful broader-than-requested results. Omit unsupported fields until wired, or reject them explicitly.
- `packages/api/src/graphql/sdl/resolvers/queries/conditionGroups.ts:170` — tag filtering can still match a private tagged condition while only requiring some unrelated public condition in the group. Put `public: true` in the same `some` as `tags: { hasSome: ... }`.
- `docs/sapience-graphql-api-redesign.md:1876` — the migration rule is right, but this PR violates it for the three root fields above.

Checks run:
- `DATABASE_URL=... pnpm --filter @sapience/api run generate-types` ✅ no generated diffs now
- targeted Vitest for cursor + outcome tests ✅
- `pnpm --filter @sapience/api run type-check` ❌ still fails on `ctx.loaders` optionality in unchanged files (`Account.ts`, `crud.ts`); looks like base/generated-type debt, but CI may still block this branch.

🔮 Much closer, but the compatibility break is still the big one.

nostradamus-agent (changes requested):
Re-review at `82a571169`: the resolver cleanup is better — generated artifacts are fresh, unsupported `QuestionFilter` fields are removed, and the `conditionGroups` private-tag bug is fixed.

Still requesting changes on compatibility: adding `@deprecated` to `conditionGroupsLegacy` / `conditionsLegacy` / `questionsLegacy` does **not** preserve the old API. Existing clients still call `conditionGroups(...)`, `conditions(...)`, and `questions(...)`; those names now have different args and return types, so their queries still fail validation. The one-release window has to keep the old field names/shapes alive while adding the new connection surface under additive names (or this needs to be explicitly treated as a breaking release).

Concrete example: a pinned query like `conditions(where: { public: { equals: true } }, take: 10) { id }` used to validate. On this PR, `conditions` no longer accepts `where`/`take` and returns `ConditionConnection!`, while the compatible shape moved to `conditionsLegacy`. Renaming is itself the breaking change.

Checks run:
- `DATABASE_URL=... pnpm --filter @sapience/api run generate-types` ✅ no generated diffs
- targeted cursor/outcome Vitest ✅
- `pnpm --filter @sapience/api run type-check` ❌ still fails on unchanged `ctx.loaders` optionality (`Account.ts`, `crud.ts`), likely base/generated-type debt but still worth watching in CI.

🔮 We’re down to the core schema-versioning call now: keep old names for a release, or mark this PR as intentionally breaking.

noahlitvin (approved):

nostradamus-agent (approved):
Approved PR2 after re-review. The remaining follow-up (#1742) covers the operator filter/keyset cursor gap; assuming that lands into this branch before merge, this is good to go.

nostradamus-agent (approved):
Re-approved after resolving the staging merge conflict in the design doc.

PR #1764, opened 2026-05-20
Add Dockerfile for ECS builds

llamalm:
## Summary

Adds a top-level \`Dockerfile\` so CI in this repo can build container images for ECS without depending on a sibling \`sapience-dx\` checkout. 

Adapted from the legacy \`sapience-dx/dockerfiles/Dockerfile.monorepo\`:

- Bumped pnpm to **10.33.4** to match this repo's \`packageManager\` field (the old one was pinned to 9.15.9).
- Package manifest list matches what actually lives under \`packages/\` here (includes \`benchmarks\`, no \`commitment-executor\`).
- Same multi-stage layer caching: workspace + manifests → install with ignored scripts → rebuild for native modules → SDK build → full source copy → re-install with scripts → Prisma generate.

## Why

CI workflows can only checkout the repo they run in. Pointing builds at \`../sapience-dx/dockerfiles/Dockerfile.monorepo\` was leftover from the local dev compose setup and doesn't work on a GitHub Actions runner without cross-repo PATs. Having a \`Dockerfile\` at the repo root is also the standard convention.

## Test plan

- [x] \`docker build -f Dockerfile -t sapience:test .\` succeeds locally on macOS arm64.


nostradamus-agent (approved):
Approved. Re-reviewed the added `.dockerignore` plus the Dockerfile; the ignore list keeps local junk/secrets/build outputs out of the context without excluding tracked files needed by the Dockerfile. Build-step emulation passed earlier and current Vercel statuses are green.

PR #1783, opened 2026-05-22
fix(markets): cursor-paginate questions infinite scroll

noahlitvin:
## Summary

- Staging \`/markets\` stopped loading more rows after the first page with \`QUERY_COMPLEXITY_EXCEEDED\` (limit 15000, actual 16361). Captured the failing request via DevTools: \`take: 40, after: null\` — the SDK was asking the server for \`first = skip + take\` from the start on every scroll.
- SDK \`fetchQuestionsPage\` now makes one constant-cost request per page (pass-through \`after\`, returns \`{ items, hasMore, endCursor }\`).
- \`useInfiniteQuestions\` switches to \`useInfiniteQuery\` keyed off the returned cursor — same pattern \`usePositions\` and \`useForecasts\` already use.

## Root cause

\`questionsConnection\` is cursor-only on the server side (the \`take\`/\`skip\` deprecated args were removed in 3d3333a19). But the SDK was emulating offset pagination on top:

\`\`\`ts
// before: skip=20 → first=40 from start; skip=40 → first=60 from start
while (collected.length < target) {
  const first = Math.min(100, target - collected.length);
  ...
}
return collected.slice(skip, target);
\`\`\`

Cost grew linearly with depth, so by page 2 the query exceeded the 15k complexity budget — and silently. The hook saw an error, didn't advance \`skip\`, and the user saw \"infinite scroll just stopped working.\"

## Sibling helpers worth following up on

\`fetchConditions\` and \`fetchConditionGroups\` in the SDK still use the same skip-walking pattern. They don't fire the bug today because their only callers (\`useConditions\`, \`useConditionGroups\`) are non-paginating \`useQuery\` calls with \`skip=0\`. But the trap is sitting there — worth deleting the walking from those helpers too (drop the \`skip\` arg, accept \`after\` instead) so nobody can stumble into the same failure mode later.

## Test plan

- [x] \`pnpm --filter @sapience/sdk run test\` — 671 pass (added cursor-passthrough + 100-cap tests)
- [x] \`pnpm --filter @sapience/app run test --run\` — 626 pass
- [x] \`pnpm --filter @sapience/app run type-check\` — clean
- [x] \`pnpm --filter @sapience/app run lint\` — clean
- [ ] Smoke on staging once merged: open \`/markets\`, scroll past page 1, confirm rows keep loading and the Network tab shows \`first=20, after=<cursor>\` instead of \`first=N, after=null\`.

### DIFF
--- a/packages/app/src/hooks/graphql/__tests__/useInfiniteQuestions.test.ts
+++ b/packages/app/src/hooks/graphql/__tests__/useInfiniteQuestions.test.ts
@@ -1,0 +1,110 @@
+import { vi, describe, it, expect, beforeEach } from 'vitest';
+import { act, renderHook, waitFor } from '@testing-library/react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import React from 'react';
+import type { QuestionType } from '../useInfiniteQuestions';
+
+const mockFetchQuestionsPage = vi.fn();
+
+vi.mock('@sapience/sdk/queries', () => ({
+  fetchQuestionsPage: (...args: unknown[]) => mockFetchQuestionsPage(...args),
+}));
+
+function createWrapper() {
+  const queryClient = new QueryClient({
+    defaultOptions: {
+      queries: { retry: false, gcTime: 0 },
+    },
+  });
+  return function Wrapper({ children }: { children: React.ReactNode }) {
+    return React.createElement(
+      QueryClientProvider,
+      { client: queryClient },
+      children
+    );
+  };
+}
+
+async function getHook() {
+  const mod = await import('../useInfiniteQuestions');
+  return mod.useInfiniteQuestions;
+}
+
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((res) => {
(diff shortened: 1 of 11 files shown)


nostradamus-agent (approved):
Reviewed the cursor-pagination changes. No blocking findings.

What I checked:
- `useInfiniteQuestions` now uses `useInfiniteQuery` with `endCursor` as `pageParam`, so scroll fetches stay `take=pageSize, after=<cursor>` instead of growing `first` from zero.
- SDK `fetchQuestionsPage` is now one cursor page for new callers, while legacy `skip`/large-`take` callers still get window-walking compatibility through capped 100-row requests.
- Same compatibility pattern for `fetchConditions` / `fetchConditionGroups`.
- Added tests cover cursor pass-through, legacy skip windows, large-take callers, and the same-frame `fetchMore` coalescing latch.

Verification: `git diff --check` passed; GitHub commit status is green for Vercel app/docs. Local vitest/typecheck could not run in this sandbox because Node/pnpm hit resource/module-linkage issues, so I did not claim local test verification.

PR #1787, opened 2026-05-23
fix vault ID

7ayd:


nostradamus-agent (approved):
LGTM. Reviewed the tiny resolver fix: `vault(id:)` now returns `null` for malformed/non-Vault IDs instead of throwing, while valid Vault global IDs still resolve. CI is green; local targeted test run was blocked by sandbox inode exhaustion, so I relied on static diff review + CI.

noahlitvin:
Superseded — the same vault tolerance fix shipped to staging in #1799. #1790 is the cleaner alternative if/when this lands on main. The extra  guard here is a no-op at runtime because  already returns null for non-vault-shaped inner ids.

PR #1790, opened 2026-05-23
fix(api): Query.vault(id:) returns null for malformed global ids

noahlitvin:
## Summary

- `Query.vault(id:)` called `fromGlobalId` directly, which throws `InvalidGlobalIdError` on inputs that don't round-trip through base64url (e.g. a raw EVM address).
- Switch to `tryFromGlobalId` so the resolver matches the already-tolerant `Query.node(id:)` path and returns `null` instead of surfacing a logged GraphQL error.
- Adds a regression test that reproduces the Sentry `malformed global id payload: 0x0000…0000` error before the fix.

## Sentry

Resolves:
> `GraphQLError: malformed global id payload: 0x0000000000000000000000000000000000000000 (decoded: �4�M4�M4�M4�M4�M4�M4�M4�M4�M4�)`

## Test plan

- [x] `pnpm --filter @sapience/api exec vitest --run src/graphql/sdl/resolvers/queries/analytics.test.ts`
- [x] type-check passes
- [ ] CI green

### DIFF
--- a/packages/api/src/graphql/sdl/resolvers/queries/analytics.test.ts
+++ b/packages/api/src/graphql/sdl/resolvers/queries/analytics.test.ts
@@ -31,7 +31,7 @@ vi.mock('../../../../lib/utils', () => ({
   getProviderForChain: vi.fn(),
 }));
 
-const { protocolStats, vaultStats } = await import('./analytics');
+const { protocolStats, vaultStats, vault } = await import('./analytics');
 
 type StatsArgs = {
   fromEpoch?: number | null;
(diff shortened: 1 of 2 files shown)


noahlitvin:
Superseded — the same  fix shipped to staging in #1799 alongside the claimsConnection/closesConnection work, so staging is already unblocked. Main is currently pre-staging-merge (the resolver doesn't even use Relay there), so this can be reopened/rebased if main ever catches up to the Relay-Node migration.

PR #1813, opened 2026-05-29
Activate event-keyed condition_group identity end-to-end

ukitta555:
## Summary

Bundles three changes (originally planned as PR A + PR B + PR C) that together switch `condition_group` identity from name-keyed to `(source='polymarket', externalEventId)`-keyed and fix the long-standing name-collapse bug. Each is unobservable or unsafe without the others, so they ship as one atomic unit.

- **API admin route** (`packages/api/src/routes/conditions.ts`, `conditionGroups.ts`): every group lookup site uses a new precedence — `(source, externalEventId)` first when the payload carries an event id, falling back to `name` when it doesn't. Legacy traffic (no `externalEventId`) goes through the exact same code path as today.
- **Keeper** (`packages/market-keeper/src/generate/{grouping,api}.ts`): includes `externalEventId` (from `market.events[0].id`) in every admin POST payload — `submitConditionGroup`, `submitCondition`, `batchCreateConditions`. Going forward, fresh Polymarket events create fresh Sapience groups instead of getting folded by display name.
- **Schema** (migration `20260529001202_drop_condition_group_name_unique`): drops `UNIQUE(name)`. Required so the keeper's per-event create can succeed when multiple Polymarket events share a display title. Replaced with a non-unique B-tree index so the legacy name-fallback stays fast. The partial `UNIQUE(source, externalEventId) WHERE externalEventId IS NOT NULL` from #1809 is now the canonical identity constraint.

## Why bundled

| If only this ships | Result |
|---|---|
| API change alone | New code path is never exercised — keeper still sends only \`groupName\`. No observable effect; no risk; also no fix. |
| Keeper change alone | API doesn't know about \`externalEventId\` → silently ignored → still name-keyed. No fix. |
| API + Keeper without dropping \`UNIQUE(name)\` | Next week's "Top US Netflix Movie This Week" event tries to create a fresh group → \`UNIQUE(name)\` violation → keeper writes break. |
| All three | Fix is atomic. New events create fresh groups. Existing data unaffected. |

## Required gating

**Run the event backfill on the target environment BEFORE merging this PR.** Without populated \`externalEventId\` on existing groups, the new event-keyed lookups all miss, so the first keeper sync after deploy creates a duplicate fresh group for every recurring event (instead of reusing the populated existing one).

\`\`\`bash
PRISMA_QUERY_TIMEOUT_MS=60000 pnpm --filter @sapience/api exec tsx \\
  scripts/backfillConditionGroupEvent.ts --apply --resolve-conflicts
\`\`\`

The backfill is idempotent and read-once-write-once; re-runs are safe.

### New shared helper

[\`packages/api/src/routes/helpers/conditionGroupIdentity.ts\`](packages/api/src/routes/helpers/conditionGroupIdentity.ts) exports:

- \`resolveOrCreateGroup({name, externalEventId?, negRiskMarketId, ...})\` — single-shot find-or-create with the documented precedence rule. Race fallback re-queries on the same key that was used for the original lookup.
- \`resolveGroupsForBatch(payloads)\` — pre-fetches by both keys in parallel (\`Map<eventId, row>\`, \`Map<name, row>\`).
- \`pickGroupForPayload(payload, maps)\` — per-item picker the batch loops call.

### Batch flows

Both \`POST /admin/conditions/batch-create\` and \`PUT /admin/conditions/batch-metadata\` are refactored from name-keyed maps to identity-keyed maps (\`event:<eid>\` or \`name:<nm>\` per item). The basket invariant and leader-takes-all logic are preserved unchanged — only the pre-fetch + group resolution is restructured.

### Keeper plumbing

- \`SapienceCondition\` and \`SapienceConditionGroup\` (types/sapience.ts) gain \`externalEventId?: string\`.
- \`MarketGroup\` (filters/volume-threshold.ts) gains \`eventId?: string\`.
- \`groupMarkets\` populates it from \`event.id\` at the same place it already sets \`title\` and \`eventSlug\`.
- \`transformToSapienceCondition\` and \`freshMetadataFor\` thread it through.
- All three admin-POST builders (\`submitConditionGroup\`, \`submitCondition\`, batch payload mapper) include it in the body.

### Backward compatibility

- Any payload without \`externalEventId\` (legacy ops scripts, the misc \`packages/market-keeper/misc/*.ts\` tools, an older keeper if rolled back) lands on the legacy name-fallback path, which is bit-identical to today's behavior.
- \`UNIQUE(name)\` drop is one-way once duplicates land. To revert: re-merge the duplicates first (operator script), then re-add the constraint.

### Failure mode

The trickiest piece is the batch-create invariant code ([\`conditions.ts:240-286\`](packages/api/src/routes/conditions.ts#L240-L286) area). The new tests cover the event-keyed admission + leader-takes-all + race fallback paths; reviewer should pay extra attention there.

## Test plan

- [x] All 515 existing API tests pass
- [x] 6 new tests added in \`conditions.test.ts\`:
  - new-group create via event-keyed path
  - reuse existing group via \`externalEventId\` even when name differs
  - legacy name-fallback regression
  - batch event-keyed new-group create
  - mixed batch (event-keyed + legacy items in the same payload)
  - batch invariant: event-keyed group rejects basket mismatch
- [x] Keeper tests pass (3 pre-existing failures in \`refresh-metadata.test.ts\` testing \`fetch.ts\` shape mismatch — not introduced by this PR, see commit \`git diff origin/staging -- packages/market-keeper/src/refresh-metadata\` which is empty)
- [ ] Migration applies cleanly on staging
- [ ] After deploy, bake ≥1 Polymarket weekly cycle to confirm a real recurring event lands in a fresh \`condition_group\` row:
  \`\`\`sql
  SELECT id, name, "externalEventId", "createdAt"
  FROM condition_group
  WHERE name = 'Top US Netflix Movie This Week'
  ORDER BY "createdAt" DESC LIMIT 5;
  -- Should show a new row created post-deploy with a distinct externalEventId.
  \`\`\`


nostradamus-agent (changes requested):
🔴 Requesting changes: metadata refresh computes `externalEventId`, but never includes it in the batch-metadata payload, so existing-condition regrouping still falls back to non-unique `groupName` after this PR drops `UNIQUE(name)`.

If Polymarket renames/retags an existing condition and `groupName` is emitted, `PUT /admin/conditions/batch-metadata` sees no `fields.externalEventId`, resolves by `name`, and `conditionGroupIdentity.resolveGroupsForBatch()` intentionally picks the newest duplicate-name row (`orderBy id desc`). That can move an existing condition onto the wrong recurring event group — basically the name-collapse bug wearing a fake mustache.

Fix: include `externalEventId` whenever a metadata update includes `groupName` / needs group resolution, or expose/read the current group event id and diff it explicitly so batch-metadata always has the event key for Polymarket-owned rows.

Inline anchor failed; finding is in `packages/market-keeper/src/generate/grouping.ts:462-471`.

nostradamus-agent (changes requested):
🔴 Still requesting changes: the fix covers `externalEventId` drift, but not the more common `groupName` drift path.

`freshMetadataFor()` now computes `externalEventId`, and `computeMetadataUpdates()` includes it in the diff key list, but the diff loop only emits fields whose value changed. If an existing condition already has the correct `conditionGroup.externalEventId` and Polymarket only changes the event title/group name, the keeper still sends a batch-metadata update with `fields: { groupName }` and no `externalEventId`.

That means `/admin/conditions/batch-metadata` still takes the legacy name-keyed path (`updateIdentityKey()` falls back to `groupName`) and can resolve the newest duplicate-name group after `UNIQUE(name)` is dropped. The current unit test actually demonstrates this: `metadata-updates.test.ts` still expects the group-name drift case to emit only `groupName`, and the run logs `changed: groupName`.

The other touched admin endpoints look structurally OK for the keeper payloads:
- `POST /admin/conditions/batch-create` uses event identity when `externalEventId` is present and only falls back to name when absent.
- `POST /admin/conditions` and `PUT /admin/conditions/:id` use `resolveOrCreateGroup()` when `groupName` is present; keeper create payloads include both `groupName` and `externalEventId`.
- `POST/PUT /admin/conditionGroups` are direct group CRUD, not ambiguous name resolution; uniqueness now correctly lives on `(source, externalEventId)`.
- `PUT /admin/conditionGroups/:id/conditions` is ID-addressed, so duplicate names do not matter.

Fix I want: whenever a metadata update emits `groupName` for a row with a fresh `externalEventId`, include `externalEventId` in the same payload even if the event id did not itself change. Add the regression test as “groupName changed + externalEventId unchanged => fields includes both”.

Checks on current head `3762e5f2068e338e5e577a801220bcde658b7394`:
- market-keeper metadata tests: 33/33 passed
- API route tests: 89/89 passed
- `@sapience/market-keeper` type-check passed
- `@sapience/api` type-check passed

PR #1814, opened 2026-05-29
Release: staging → main

ukitta555:
## Summary

Promotes the net `staging` state to `main`. The GraphQL surface work that landed on staging earlier in the cycle was rolled back by #1802 in prep for a `/v2` endpoint, so it is **not** in this diff. What's actually here:

- **condition_group event identity** — add `source` + `externalEventId` columns with a partial unique index `(source, externalEventId) WHERE externalEventId IS NOT NULL`, plus `scripts/backfillConditionGroupEvent.ts` to populate them. Existing `UNIQUE(name)` stays in place; the keeper still uses name-keyed lookups until a follow-up flips it (#1809, #1812).
- **negRisk metadata on condition groups** — new `negRiskMarketId` column, `IDX_condition_group_neg_risk_market_id` partial index, GraphQL `ConditionGroup.negRisk: Boolean!` derived from `negRiskMarketId IS NOT NULL`, backfill script, and market-keeper pipeline support (#1727).
- **Question-sort partial indexes on `condition`** — `createdAt`, plus `volume{1h,4h,24h,7d}` and `volumeFiltered{1h,4h,24h,7d}` filtered on `public = true`, closing the seq-scan gap on the ungrouped branch of `questionsConnection`.
- **Discord alerts for secondary trades** — new `SecondaryTradeAlertData` path in `discordAlert.ts` and a hook in `secondaryMarketIndexer` (#1747).
- **API runtime** — concurrency limiter rework, new inflight registry + dump (`inflightDump.ts`, `inflightRegistry.ts`), middleware test coverage, and `deprecationTelemetry` lib.
- **API services** — new `accountStats` and `forecastNormalization` services (+ tests).
- **API routes** — referrals route added; conditions route gets AP-style title-case normalization for Polymarket tag labels.
- **market-keeper** — restore `refresh-imminent-tag` after the #1802 revert (#1805); add LLM endTime smoke script and silence per-market debug logs (#1807); various LLM/grouping/tags/templated updates; negRisk submit path.
- **App** — remove example combos refresh button (#1770); minor `Comments`, `ProfileQuickMetrics`, `CommandMenu`, og-helpers, and `useTradeSettledNotifications` tweaks.
- **Infra/CI/tests** — add root `Dockerfile` + `.dockerignore`, `.npmrc` tweak, `railway.toml` updates, deepen CI checkout so paths-filter can diff without fetching (#1801), and update the API contract fixture to use the current `attestation.prediction` column so contract tests seed correctly.

## Test plan

[...]

PR #1802, opened 2026-05-25
revert: roll graphql surface back to main (prep for /v2 endpoint)

noahlitvin:
## Summary

Roll the public GraphQL surface on `/` back to exactly main so promoting staging → main doesn't break the external consumer. The cleaner Relay-shaped surface we've been iterating on will land in a follow-up PR as a separate `/v2` endpoint.

This trades the in-place rename / shim approach (which keeps producing edge cases like `Condition.predictions` quietly changing return type from `[LegacyPrediction!]!` → `PredictionConnection!`) for a clean endpoint split: `/` is a contract, `/v2` is the product.

## What's reverted to match main exactly

- `packages/api/schema.graphql` + `packages/api/src/graphql/**` (schema, resolvers, codegen)
- `packages/sdk/types/graphql.ts` + `packages/sdk/queries/**`
- `packages/app/**` — every FE file that hit the schema (hooks/graphql, lib/data, all call sites in components and routes)
- `packages/docs/**` — builder-guide samples used `*Connection`
- `packages/relayer/src/loadtest/**` — referenced `accountStatsLeaderboardPage`
- Contract tests + introspection snapshots
- `workers/indexers/easIndexer`, `predictionMarketEscrowIndexer` — touched the renamed `forecast` column
- `services/scoringService`, `services/protocolStats/snapshots`
- Restored `services/positionPnL`, `services/predictionNormalization` (main's resolvers depend on them; they were deleted on staging)
- `prisma/schema.prisma` reverted; the `rename_attestation_prediction_to_forecast` migration is dropped (main reads from the legacy column name)

## What stays from staging (kept; promotable as-is)

- Infra: `Dockerfile`, `.dockerignore`, `.npmrc`, `ci.yml` fetch-depth fix, `railway.toml` builder config
- API runtime hardening: `runtime/concurrencyLimiter`, `runtime/middleware`, `runtime/inflightRegistry`, `runtime/inflightDump`; `core/server` socket timeouts; `core/db` pool work
- `lib/deprecationTelemetry` (orphan but harmless; /v2 will wire it back up)
- API HTTP routes: `routes/referrals` (new), `routes/conditions` changes
- API orphan services: `services/accountStats`, `services/forecastNormalization` (dead code on `/`; /v2 picks them up)
- Prisma index migrations: `add_questions_sort_indexes`, `add_remaining_volume_indexes` (additive, `IF NOT EXISTS` guarded)
[...]

PR #1821, opened 2026-05-30
Release: staging → main

ukitta555:
Promotes the current staging state to main. Covers everything merged to staging since the last release (PR #1814).

### Event-keyed condition_group identity (#1813)
Switches \`condition_group\` identity from name-keyed to \`(source='polymarket', externalEventId)\`-keyed end-to-end:
- **Schema**: drops \`UNIQUE(name)\` (replaced with a non-unique index); the partial unique on \`(source, externalEventId)\` from #1809 is now the canonical identity constraint.
- **API admin route**: \`POST/PUT /admin/conditions*\` and \`POST/PUT /admin/conditionGroups*\` look up groups by \`(source, externalEventId)\` first, falling back to name. Refactored four lookup sites + introduced \`packages/api/src/routes/helpers/conditionGroupIdentity.ts\`.
- **Keeper**: includes \`externalEventId\` (from \`market.events[0].id\`) in every admin POST and in batch-metadata payloads. Diff-loop now force-includes \`externalEventId\` whenever \`groupName\` is emitted (so route never falls through to name-fallback for ambiguous lookups).
- **GraphQL**: \`ConditionGroup.externalEventId\` exposed; keeper consumes it via the existing GraphQL refresh query.
- **Tests**: +14 new tests in \`conditions.test.ts\` and \`metadata-updates.test.ts\` covering the new identity precedence, the metadata-diff force-include for groupName, and regression coverage for the legacy name-fallback path.

### negRisk basket fixes
- #1817 — closes basket gaps in the API + market-keeper, and fixes a prod prompt-log leak.
- #1818 — refresh-metadata mixed-basket bug, plus wiring GraphQL operation names.

### Other improvements
- #1816 — \`combo-bingo.vercel.app\` CORS allow.
- #1811 — staging GraphQL client compatibility fix.
- LLM enrichment / openrouter / prompt updates (changes across \`packages/market-keeper/src/llm/*\`).
- \`refresh-imminent-tag\` restored after the #1802 revert (#1805).

## Operator action required (before merging this PR to main)

The event-keyed identity work in #1813 ships a new migration (\`drop_condition_group_name_unique\`) and switches the keeper to send \`externalEventId\`. Prod data was already prepared during the staging rollout via the operator scripts:

- ~35,005 of 35,024 condition_group rows now have \`externalEventId\` populated.
- 19 NULL rows are the legacy / unknown-to-Polymarket cohort; they use the name-fallback path post-deploy.
- Zero outstanding \`group collisions (skipped)\` after merge-pass cleanup.

If you need to re-verify pre-deploy, see PR #1820 (operator scripts) for the dry-run + apply pipeline.

## Test plan

- [ ] CI green on the merge commit
- [ ] Prisma migrations apply cleanly against prod (new migration: \`drop_condition_group_name_unique\`)
- [ ] After deploy, monitor admin-route logs for unexpected 400s for the first ~30 min — none expected, since the new path activates only when keeper sends \`externalEventId\` (which it now does)
- [ ] Spot-check via SQL that the next recurring weekly event (e.g. \"Top US Netflix Movie This Week\") creates a new \`condition_group\` row with its own \`externalEventId\` instead of folding into the existing group:
  \`\`\`sql
  SELECT id, name, \"externalEventId\", \"createdAt\"
  FROM condition_group
  WHERE name = 'Top US Netflix Movie This Week'
  ORDER BY \"createdAt\" DESC LIMIT 5;
  \`\`\`
- [ ] Confirm refresh-metadata sync runs cleanly post-deploy (basket invariant should be a no-op since externalEventId always travels alongside groupName now)


nostradamus-agent (approved):
LGTM — staging → main promotion approved.

PR #1831, opened 2026-06-09
Release: staging → main

noahlitvin:
## Release: staging → main

Promotes the current `staging` branch to `main`. 30 commits, spanning API hardening, event-keyed condition-group identity, the new market-keeper endTime labeling tool, and UI hotfix back-ports.

### Highlights

**Event-keyed condition_group identity**
- Activate event-keyed `condition_group` identity end-to-end (#1813)
- Add event identity columns to `condition_group` (#1809); event-id backfill tool (#1812)
- Operator scripts for event-id backfill collision pre-flight + dedup (#1820)
- negRisk metadata on condition groups (#1727); close negRisk basket gaps + fix prod prompt-log leak (#1817)

**API**
- Attribute claim PnL via `pickConfigId` so `accountPnl` is not always 0 (#1827)
- Add `claimsConnection` / `closesConnection`, harden `vault(id:)` (#1799); singular by-id queries (#1797); un-deprecate approved queries (#1796)
- Restore natural id fields on 5 types (breaking-change rollback) (#1800); roll graphql surface back to main to prep for /v2 endpoint (#1802)
- Allow `combo-bingo.vercel.app` CORS (#1816); restore GraphQL client compatibility (#1811)
- Contract snapshot + introspection/UTF-8 fixes

**market-keeper**
- endTime labeling tool + dataset + regex extractor (#1824)
- refresh-metadata mixed-basket bug fix + wire gql operation names (#1818)
- Restore refresh-imminent-tag (#1805); LLM endTime smoke + quieter logs (#1807)

**App / UI**
- Port main-only UI hotfixes back to staging (#1829)
- Remove example combos refresh button (#1770); Discord alerts for secondary trade executions (#1747)

**CI / infra**
- Deepen checkout so paths-filter can diff without fetching (#1801)
- `fix(sdk)`: make forecast order variables non-null (#1791)


packages/app/src/lib/ens/avatar.ts
```diff
@@ -113,8 +113,14 @@ function parseEnsAvatarCaip(
 }
 
 async function fetchJson<T = unknown>(url: string): Promise<T | null> {
+  // SSRF guard: token metadata URLs come from attacker-controllable on-chain
+  // tokenURI/uri values, so host-check every URL before fetching — not just the
+  // final image URL. Blocks private/link-local/cloud-metadata hosts and
+  // non-http(s) schemes.
+  const safe = sanitizeAvatarUrl(url);
```
nostradamus-agent: Blocking: this now guards metadata fetches, but the guard underneath is still regex-only and misses internal IPv6/link-local forms (`[fe80::1]`, `[fc00::1]`, IPv4-mapped IPv6, etc.), plus DNS names resolving to private ranges. Since token metadata URLs are attacker-controlled, use real IP/CIDR parsing after normalization and either resolve/block private DNS targets or route through a constrained fetch proxy/allowlist.

packages/relayer/src/secondaryMarketHandlers.ts
```diff
@@ -251,6 +252,22 @@ export async function handleSecondaryBidSubmit(
     return true;
   }
 
+  // Per-connection cap — one connection cannot monopolize a listing's bid
+  // slots. Offline validation passes legitimate smart-account (session-key)
+  // bids through as 'unverified' alongside forged ones, so without this cap a
+  // single connection could fill the listing's bid cap and lock out real
+  // buyers.
+  const bidCountKey = `secondary:${payload.auctionId}`;
```
nostradamus-agent: Blocking: the cap check happens after `validateSecondaryBid`, so an already-over-cap connection can still force signature validation work on every message. Those over-cap rejects return `false`, so they also do not count toward the validation-failure disconnect path. Move this cap before signature validation after only the minimal listing/payload-key lookup needed to compute the counter.

nostradamus-agent (changes requested):
🔮 Review verdict: request changes.

Blocking:
- `test-api` is red because generated GraphQL SDK types are stale. CI regenerated `packages/sdk/types/graphql.ts` and found doc/comment diffs for `Claim.predictionId` / `claims`, so the generated file needs to be committed.
- ENS avatar SSRF guard still has bypasses for IPv6 internal/link-local hosts (and likely DNS-to-private rebinding). This PR is explicitly hardening SSRF, so partial host regexes are not enough.
- Secondary-market bid cap runs after signature validation, so an over-cap connection can keep forcing EIP-712 verification work and those rejections do not increment the validation-failure disconnect counter.

Non-blocking notes:
- The market-keeper labeler server has unauthenticated write endpoints and binds with `server.listen(PORT)`; probably fine as local tooling, but please don’t expose it.
- The `bestEndTime` “production cascade” helper is used by labeler scoring, not the actual production grouping path for sports `gameStartTime`; rename/comment if that was intentional.

Validation I checked:
- GitHub checks: all Actions green except `test-api`; Railway/Vercel statuses mostly green with two Railway cron/smoke statuses still pending at review time.
- Focused API/indexer tests via sub-audit: 78 passed.
- Market-keeper typecheck + focused extractor/endTime tests via sub-audit: passed.

PR #1848, opened 2026-06-11
fix(app): raise vault deposit cap to 20k

sunnyvempati:
## Summary
- `DEPOSIT_CAP` in `VaultsPageContent.tsx` was hardcoded at 10000, gating deposits ("Exceeds Vault Capacity") and the capacity meter.
- Raised to 20000 to match the keeper-side cap just activated on prod `vault-deposit-manager` (`MAX_ASSETS=20000000000000000000000` — note that env var had been unset, i.e. the keeper cap was previously not enforced at all).
- Follow-up worth considering: derive this from config/API instead of a UI constant so the two caps can't drift again.


sunnyvempati:
Recreating off main to avoid staging-history conflicts.

PR #1857, opened 2026-06-12
fix(sdk): buffer vault deposit share requests

noahlitvin:
## Summary
- Add a 10 bps safety buffer to SDK vault deposit share calculations
- Deposit requests now ask for slightly fewer shares to tolerate small PPS/shareValue drift before vault-claimer processing
- Leave withdrawal expected-asset calculation unchanged

## Why
Vault-claimer rejects deposits when `assets / requestedShares` is below current share value. We observed skipped deposits where the UI/SDK PPS was only a few bps stale, causing requested shares to be too high.

## Test Plan
- pnpm --filter @sapience/sdk run build:lib
- pnpm --filter @sapience/sdk run type-check
- pnpm --filter @sapience/sdk run lint
- pnpm --filter @sapience/sdk run test -- onchain/__tests__/vault.test.ts

### DIFF
--- a/packages/sdk/onchain/__tests__/vault.test.ts
+++ b/packages/sdk/onchain/__tests__/vault.test.ts
@@ -1,5 +1,5 @@
 import { describe, test, expect } from 'vitest';
-import { parseAbi } from 'viem';
+import { parseAbi, decodeFunctionData } from 'viem';
 import type { Abi } from 'abitype';
 import type { Address } from 'viem';
 import {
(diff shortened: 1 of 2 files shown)


nostradamus-agent (approved):
🔮 Reviewed and approved.

No blocking findings. The SDK now requests ~10 bps fewer deposit shares, which matches the vault-claimer tolerance goal without touching withdrawal math.

Verified locally:
- `pnpm --filter @sapience/sdk run lint`
- `pnpm --filter @sapience/sdk run test -- onchain/__tests__/vault.test.ts`

PR #1869, opened 2026-06-15
feat(app/sdk/api): migrate vault stats charts to /v2/graphql

noahlitvin:
Migrates the vault dashboard's stats/charts off the v1 `protocolStats(vaultAddress:)` endpoint (`/graphql`) onto the v2 `vault(address:, chainId:).statsHistory` surface (`/v2/graphql`), plus the one small v2 API addition that unblocks it.

## API
- Add `unredeemedClaim: BigInt!` to `type VaultStat` (`schema/schema.graphql`) — wUSDe owed to the vault on resolved-but-not-yet-redeemed winning sides, net of what it has already claimed.
- `mapVaultStat` in `resolvers/Vault.ts` now returns `unredeemedClaim: BigInt(row.vaultUnredeemedClaim ?? '0')` (the column was already read inside the existing `cumulativePnl` roll-up).
- Regenerated the committed `schema.v2.graphql` snapshot and `sdk/types/graphql.v2.ts` via `pnpm --filter @sapience/api run generate-types` (no hand-edits to generated files).
- Test: extended `Vault.test.ts` to assert `unredeemedClaim` maps from `vaultUnredeemedClaim`.

## SDK
- New `queries/vault.ts`: `fetchVaultStats(address, chainId?, first=1000)` over `graphqlRequestV2`, querying `vault(address:, chainId:).statsHistory` and selecting `timestamp/balance/deployedCollateral/undeployedCollateral/cumulativePnl/unredeemedClaim`. `statsHistory` defaults to TIMESTAMP DESC, so the mapper sorts ascending (mirrors `protocol.ts`). Returns `[]` when the vault is unknown (null `vault`). Exported from `queries/index.ts`.
- New test `queries/__tests__/vault.test.ts` (mocks the transport; asserts document/vars, ascending sort, BigInt-scalar normalization, empty-on-null).
- Removed the now-orphaned v1 `queries/analytics.ts` (`fetchProtocolStats`, `GET_PROTOCOL_STATS`, `ProtocolStat`, and the unused `fetchOpenInterestByCategory` / `fetchOpenInterestByTimeToResolution` + their types/queries) after confirming zero live consumers.

## App
- New adapter `lib/adapters/vaultStat.ts` (+ sibling test) mapping a v2 `VaultStat` node to the chart point shape, mirroring the `pickConfig` adapter pattern. **TVL = `balance + deployedCollateral + unredeemedClaim`** (÷1e18), **PnL = `cumulativePnl`** (÷1e18). This replaces v1's `vaultBalance + chain-wide escrowBalance`, which over-counted.
- `vaultPnlChartUtils.ts` now consumes the adapted `{ timestamp, pnl, tvl }` point directly (dropped `mapProtocolStatToPoint`).
- `useProtocolStats` → `useVaultStats(vaultAddress?)` (v2-backed, `enabled`-gated, `DEFAULT_CHAIN_ID`). `VaultPnlChart.tsx` and `VaultsPageContent.tsx` rewired; multi-vault tab behavior (core/options/edge each query their own `vault(address:)`) preserved.
- `VaultsPageContent` yield calc: per-vault `deployedCollateral` replaces `vaultDeployed`; protocol-wide TVL now reads `useProtocolAnalytics().stats.totalValueLocked` (server-computed) instead of reconstructing from per-vault `escrowBalance + vaultAvailableAssets`.
- The live "now" headline TVL still comes from `usePassiveLiquidityVault` (RPC) — left intact. The chart plots recorded snapshots only, same as v1 (v1 did not append a live candle for the vault chart).
- Removed orphaned `getProtocolTvlWei` (only test-referenced; depended on v1 `ProtocolStat`) and its test.

## What was intentionally left
The v1 transport (`graphqlRequest` / `createGraphQLClient` / `getGraphQLEndpoint`, the `getGraphQLEndpoint` re-export in `og/_forecast-helpers.ts`, and the v1 `graphqlEndpoint` setting in `SettingsContext`) stays — still live in Header, ConnectDialog, referrals hooks, and `usePositions`. Only the vault-stats-specific v1 pieces were dropped.

## Verification (all green)
- `@sapience/sdk` build:lib ✅, `pnpm --filter @sapience/sdk run test` → 721 passed
- `@sapience/api` prisma:generate ✅, generate-types ✅, `run test` → 804 passed (72 files)
- `@sapience/app` `run test` → 746 passed (72 files); `type-check` ✅; eslint on touched files ✅
- `@sapience/sdk` / `@sapience/api` type-check ✅


nostradamus-agent (changes requested):
NO-SHIP — two blockers in the v2 vault stats migration:

- `ALL` history is silently truncated: the client asks `statsHistory(first: 1000)` but the v2 resolver clamps `statsHistory` to 365 and the SDK query does not request/page through `pageInfo`. Result: the vault PnL chart can only show ~365 snapshots after this migration.
- Legacy vault history gets dropped: the new path routes charts through `vault(address).statsHistory`, whose resolver filters by the current primary vault address only. The old v1 path included configured legacy vault addresses, so redeployed vaults lose historical chart data here.

Suggested fix: page `statsHistory` until complete or raise the limit for this dashboard use case, and make `Vault.statsHistory` include current + legacy addresses with timestamp dedupe/current-primary preference to match v1 behavior.

Verification: reviewed incremental diff against `4aefe56d1dca...origin/pr-1869`. Targeted tests were not run in my local review worktree because `pnpm` was unavailable there.

noahlitvin:
Both blockers addressed.

1. **Legacy addresses** — `Vault.statsHistory` now queries the vault's full address set (current primary + every demoted-to-legacy address, via `getConfiguredVaults`) and dedupes by timestamp preferring the current-primary row, matching v1 `protocolStats`. Dedupe runs before pagination so cursor pages stay correct. Regression test added.
2. **Truncation** — `fetchVaultStats` now selects `pageInfo` and pages `statsHistory` until exhausted (the resolver's per-page cap stands), so vaults older than the cap keep their earliest chart history. Multi-page accumulation test added.

API v2 resolver suite (226) + SDK + app + type-checks all green.

PR #1870, opened 2026-06-15
feat(app): migrate position balances to /v2/graphql positions connection

noahlitvin:
## Summary

Migrates the two position-balance hooks off v1's skip/take `positionsPage` onto the v2 Relay `positions` connection. **Consumer-only** — the v2 `positions` surface already exists and synthesizes rows with v1 parity, so no API/schema changes.

- `usePositionBalances` (by holder) and `usePositionBalancesByConditionId` (by condition) now go through `graphqlRequestV2` + the shared `useCursorPagination` helper (forward cursor paging on `pageInfo`, not offset/`hasMore`).
- New `lib/adapters/position.ts` centralizes the v1→v2 renames in one place:
  - `token` → `tokenAddress`
  - `side` → `isPredictorToken` (`'PREDICTOR'`)
  - `realizedPnl` → `realizedPnL`
  - surfaces v2's explicit `status` (OPEN/SOLD)
  - delegates `pickConfig` to the existing `toPickConfigData`
  - normalizes BigInt scalars (string|number) → decimal strings
- `PositionsTable` now detects sold rows via `status === 'SOLD'` instead of parsing the (now opaque) Relay row id. The `-sell-` id check stays as a fallback.
- Removes the v1 `positionsPage` documents, the orphaned `PICK_CONFIG_FRAGMENT` / `PICK_CONDITION_FRAGMENT` / `PositionBalancePage`, and the v1 `graphqlRequest` import from this module.

## Parity notes
- The v2 `positions` connection synthesizes the same OPEN/SOLD rows as v1 (per the schema's documented G5 parity), including the case where a page's rows all synthesize away but `pageInfo.hasNextPage` stays `true` with an advancing `endCursor`. The cursor helper pages on `pageInfo`, so this is preserved — covered by a test.
- `PositionsTable` is the only consumer of these two hooks.

## v1 transport
This removes `usePositions`' v1 `graphqlRequest` usage. The v1 transport itself stays (still used by Header, ConnectDialog, and the referral hooks) — that's the next migration.

## Tests
- `lib/adapters/position.test.ts` — adapter mapping (renames, status, BigInt normalization, null handling, pickConfig delegation, selection set).
- `hooks/graphql/__tests__/usePositions.test.ts` — rewritten for the v2 connection: cursor pagination, the empty-page-with-advancing-cursor case, filter/`first`/`after` variables, disabled-until-arg.
- Full app suite: **757 passed**. Type-check + lint clean.

### DIFF
--- a/packages/app/src/components/positions/PositionsTable.tsx
+++ b/packages/app/src/components/positions/PositionsTable.tsx
@@ -124,9 +124,11 @@ function PositionRow({
     ((isPredictorToken && result === 'COUNTERPARTY_WINS') ||
       (!isPredictorToken && result === 'PREDICTOR_WINS'));
 
-  // Synthetic sell rows have ids like `${dbId}-sell-${tradeHash}`; the open
-  // / parent row is just `${dbId}`.
-  const isSoldRow = position.id.includes('-sell-');
+  // v2 surfaces an explicit SOLD discriminator. (v1 encoded it in the id shape
+  // `${dbId}-sell-${tradeHash}`; the id fallback keeps any pre-migration row
+  // working.)
+  const isSoldRow =
+    position.status === 'SOLD' || position.id.includes('-sell-');
   const isClosed = !isResolved && BigInt(position.balance) === 0n;
   const realizedPnLFormatted =
     position.realizedPnL != null
--- a/packages/app/src/hooks/graphql/__tests__/usePositions.test.ts
+++ b/packages/app/src/hooks/graphql/__tests__/usePositions.test.ts
@@ -3,10 +3,12 @@ import { act, renderHook, waitFor } from '@testing-library/react';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import React from 'react';
 
-const mockGraphqlRequest = vi.fn();
+const mockGraphqlRequestV2 = vi.fn();
 
+// The position hooks page via `useCursorPagination`, which issues its requests
+// through `graphqlRequestV2`. Mock that single transport export.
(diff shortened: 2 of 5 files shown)


nostradamus-agent (changes requested):
NO-SHIP — v2 position mapping drops `pickConfig.predictionId`, which breaks claim/permalink consumers.

`POSITION_V2_FIELDS` selects `pickConfig` but not `Position.prediction { predictionId }`, and `toPositionBalance()` maps `pickConfig` via `toPickConfigData()` without backfilling the prediction id. Existing UI still needs it: `PositionsTable` returns early from `handleClaim` when `pickConfig?.predictionId` is missing, so resolved winning positions can show CLAIM but clicking becomes a silent no-op.

Suggested fix: query `prediction { predictionId }`, extend the v2 node type, set `pickConfig.predictionId = node.prediction?.predictionId ?? null`, and add a regression test for v2-mapped positions preserving it.

Verification: reviewed incremental diff against `4aefe56d1dca...origin/pr-1870`. Targeted tests were not run in my local review worktree because dependencies were missing (`vitest: not found`).

noahlitvin:
Addressed — `pickConfig.predictionId` is now preserved on v2-mapped positions.

`POSITION_V2_FIELDS` selects `prediction { predictionId }`, `PositionV2Node` carries it, and `toPositionBalance` backfills `pickConfig.predictionId = node.prediction?.predictionId ?? null`. So `handleClaim` and the prediction permalink work again. Added regression tests for both the present and null cases. Full app suite + type-check green.

PR #1871, opened 2026-06-15
feat(api/app): move referral reads off v1 GraphQL onto REST

noahlitvin:
## Summary

v2 GraphQL **deliberately does not expose referral data** (the v2 `Account` resolver says so: *"Referral data lives on the REST surface (`/referrals/*`, `/admin/referrals/*`); v2 GraphQL does not expose it"*). The referral *mutations* already used REST; this PR moves the remaining referral **reads** off v1 `graphqlRequest`.

### API
- **New `GET /referrals/users/:address`** — per-user referral status (`refCodeHash`, `maxReferrals`, presence of `referredBy` / `referredByCode` via FK ids) plus the user's own `referrals`. Public (referral data is attribution, not a credential). A missing user returns the empty/no-referral shape at **200** (not 404) so callers can read it unconditionally. This is the one piece of backend work — the user-facing reads had no REST endpoint before.

### App
- `useReferrals`: `useUserReferrals` now fetches the new endpoint via a shared `fetchUserReferralStatus` helper instead of the `user(where:)` GraphQL query (kept the `{ user }` envelope the dashboard reads).
- `Header` / `ConnectDialog`: the invite-gate status check uses `fetchUserReferralStatus` instead of inline `graphqlRequest`.
- `useAdminReferralCodes`: the two admin reads now hit the **existing** `GET /referrals/codes` and `GET /referrals/codes/:id` (single request each — the server caps the list and the admin UI does not paginate), replacing the paginated `referralCodes` GraphQL queries. Mutations were already REST.
- Removes every `graphqlRequest` import from the referral hooks + layout components.

## v1 transport
Combined with the positions PR, this removes the **last** frontend `graphqlRequest` consumers. After both land (and the vault-stats PR), the v1 GraphQL transport in the SDK can be deleted — left as a follow-up since SDK `analytics.ts` still uses it until the vault PR merges.

## Tests
- API: `GET /referrals/users/:address` — unknown-user empty shape (200, lowercased lookup), full status + referrals mapping, null `referredBy`/`referredByCode`, 500 path. **30 route tests pass.**
- App: rewrote `useAdminReferralCodes.test.ts` for the single-request REST path (was testing the removed GraphQL cursor pagination); new `useReferrals.test.ts` covers `fetchUserReferralStatus` + the `useUserReferrals` envelope mapping. **Full app suite: 749 passed.** Type-check + lint clean (api + app).

### DIFF
--- a/packages/api/src/routes/referrals.test.ts
+++ b/packages/api/src/routes/referrals.test.ts
@@ -694,3 +694,80 @@ describe('GET /referrals/codes/:id', () => {
     });
   });
 });
+
+describe('GET /referrals/users/:address', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('returns the no-referral shape (200) for an unknown user', async () => {
+    mockPrisma.user.findUnique.mockResolvedValue(null);
+
+    const res = await request(app).get('/referrals/users/0xABC');
+
+    expect(res.status).toBe(200);
+    expect(res.body).toEqual({
+      address: '0xabc',
+      refCodeHash: null,
+      maxReferrals: 0,
+      referredBy: null,
+      referredByCode: null,
+      referrals: [],
+    });
+    // address is normalized to lowercase for the lookup
+    expect(mockPrisma.user.findUnique).toHaveBeenCalledWith(
+      expect.objectContaining({ where: { address: '0xabc' } })
+    );
+  });
+
+  it('maps status + referrals, exposing FK ids for referredBy/referredByCode', async () => {
+    mockPrisma.user.findUnique.mockResolvedValue({
+      address: '0xabc',
+      refCodeHash: '0xhash',
+      maxReferrals: 5,
+      referredById: 11,
+      referredByCodeId: 22,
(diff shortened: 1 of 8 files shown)


nostradamus-agent (approved):
SHIP — referral reads moving off v1 GraphQL onto REST looks clean.

Checked the new public user-referral status route, app hook migration, invite-gate consumers, and admin-code reads. Shape preservation is sane: the user hook still returns the existing `{ user }` envelope, missing users return a stable empty 200, and mutations remain on signed admin REST.

Verification:
- `corepack pnpm --filter @sapience/app exec vitest run src/hooks/referrals/__tests__/useReferrals.test.ts src/hooks/referrals/__tests__/useAdminReferralCodes.test.ts` — passed, 10 tests.
- `corepack pnpm --filter @sapience/api exec vitest run src/routes/referrals.test.ts` — passed, 30 tests.
- `corepack pnpm --filter @sapience/app exec tsc --noEmit` — passed.

API `tsc --noEmit` still needs generated artifacts in this environment; not attributable to this PR.

PR #1875, opened 2026-06-16
feat(bingo): sponsorship — house-funded first card via OnboardingSponsor

7ayd:


packages/bingo/server/handler.ts
```diff
@@ -566,6 +611,11 @@ export async function handleApi(
     const cells = drawCells(pool.conditions, submission.seed);
     const stakePerLineWei =
       BigInt(submission.cardPriceWei) / BigInt(LINES_PER_CARD);
+    // House-fund this line iff the player's budget still covers its stake.
+    const sponsor =
+      sponsorRemaining >= stakePerLineWei
```
nostradamus-agent: 🔴 **Blocker:** this per-line decision only checks `remainingBudget >= stakePerLineWei`, which bypasses the whole-card eligibility rule in `sponsorEligibility`. A wallet with a partial leftover budget (the PR explicitly calls out `/app` 1 USDe / E11) is not eligible for a sponsored 10 USDe card, but every `/api/card/line` request will still set `predictorSponsor` as long as remaining covers one line. Since the client can fire the 10 line requests concurrently, they can all read the same partial remaining budget and attempt sponsored mints; only the first can consume it, and the rest can revert. It can also unexpectedly house-fund lines for a normal deposit card.

Fail closed here unless the card has a full bingo-card sponsorship budget — e.g. read `budgets(player)` and require `allocated >= submission.cardPriceWei` plus `remaining >= stakePerLineWei`, or persist/encode the sponsored-card decision at submit time.

packages/bingo/server/handler.ts
```diff
@@ -490,7 +520,17 @@ export async function handleApi(
...
       }),
-      prepareCollateral(network, sessionClient, player, price, undefined, true),
+      // Sponsored: the house funds the stake, so there's nothing to wrap —
+      // grant the budget (awaited to confirmation, so fundMint sees it when the
+      // 10 lines run) AND send ONE enable-only session op (a 0-wei prep) so the
+      // fresh session key is enabled serially, not raced by the line mints.
+      // Otherwise (deposit path) wrap + approve the whole card up front.
+      sponsored
+        ? Promise.all([
+            grantBudgetIfNew(network, player),
```
nostradamus-agent: ⚠️ This should fail closed for a new sponsored submit when `grantBudgetIfNew` returns `null` because the grant tx failed or the bankroll/read changed. Right now the receipt mint still succeeds in parallel, but the card may have no sponsorship budget and no upfront collateral prep, so line funding fails later after the card is already locked. Existing full-budget wallets are fine, but new-wallet grant failure needs to return a 409/502 before/without accepting the sponsored submit.

nostradamus-agent (changes requested):
🔮 Review verdict: **changes requested**.

I found a blocker in the sponsorship funding decision: `POST /api/card/line` sponsors lines based only on `remainingBudget >= stakePerLine`. That re-admits the exact partial-budget case the PR's policy/test excludes for whole-card eligibility (e.g. a 1 USDe `/app` budget on a 10 USDe bingo card), and can also unexpectedly sponsor non-sponsored/deposit cards. With the client firing line mints concurrently, each request can read the same 1 USDe remaining and all try sponsored mints; one succeeds and the rest can revert once the contract budget is consumed.

The line-funding path needs to prove this card actually has a full bingo-card sponsorship budget (for example read `budgets(player)` and require `allocated >= cardPriceWei`, plus `remaining >= stakePerLineWei`, or persist/encode the sponsored-card decision) before threading `predictorSponsor`.

Also worth tightening: a sponsored submit should fail closed if a new-wallet grant fails and no existing full-card budget is available; otherwise the receipt can be minted for a card the backend can’t fund.

Verification run locally:
- `pnpm install --frozen-lockfile --prefer-offline` ✅
- `ADMIN_TOKEN=… SERVER_SECRET=… MINTER_PRIVATE_KEY=… pnpm --filter @sapience/bingo test` ✅ (36 tests)
- `pnpm --filter @sapience/bingo run type-check` ✅
- `pnpm --filter @sapience/bingo run build` ✅

Meta: PR is currently draft and GitHub reports `mergeable_state=dirty`, so it also needs conflict resolution before this can ship.

nostradamus-agent (approved):
🔮 Re-review verdict: **approved**.

The blocker I called out is fixed: line sponsorship now derives from card price + `budgets(player).allocated/used`, not leftover budget alone, so partial `/app` budgets no longer accidentally sponsor bingo lines. Sponsored submit also now fails closed via `ensureSponsoredBudget()` before minting the receipt, so a grant failure won’t lock an unfundable card.

Verification run locally:
- `pnpm install --frozen-lockfile --prefer-offline` ✅
- `pnpm --filter @sapience/bingo test` ✅ (55 tests)
- `pnpm --filter @sapience/bingo run type-check` ✅
- `pnpm --filter @sapience/bingo run build` ✅

Non-blocking note: PR is still marked draft, but mergeability is now clean.

nostradamus-agent:
🔮 Review verdict: request changes. Two sponsorship invariants are still fuzzy, plus the admin SIWE login needs origin/chain binding.

Blockers:
- Sponsored card identity is inferred later from `cardPriceWei == 10 USDe` plus live grant state. Since non-sponsored submissions can also use that price, a normal paid card can become sponsor-funded after a grant exists. Persist/stamp sponsorship at submit time, or make the sponsored price/receipt path unambiguous.
- `getSponsoredLineContext` returns `null` both for ‘not sponsored’ and ‘sponsored but budget exhausted’. The caller treats `null` as normal player-funded flow, so a sponsored receipt can silently fall through to spending player collateral when budget is depleted/contended. Sponsored-but-unspendable should fail closed/retry, not become paid.
- Admin SIWE verifies the signature but does not enforce expected `domain`, `uri`, or `chainId` from the parsed message. Network-scoped bearer tokens help after login, but the wallet signed message itself is still replayable across origins/chains if an admin signs it.

Verification: reviewed diff against `origin/bingo`; `git diff --check origin/bingo...HEAD` passed. Could not run Bingo tsc/tests locally because the sandbox is inode-exhausted (`ENOSPC`) while installing deps.

nostradamus-agent (changes requested):
🔮 Formal verdict: request changes. Details in my top-level review comment: sponsorship identity/fail-closed behavior and SIWE origin/chain binding need fixes before this ships.

nostradamus-agent (approved):
🔮 Follow-up review: approved.

The pushed delta fixes the remaining SIWE blocker: admin login now requires the signed message chainId to match the selected Bingo network and the SIWE domain to match the request host; the same values are covered in the added adminAuth tests.

I’m treating the 10-USDe card-price/sponsorship ambiguity as accepted operational policy per Zayd’s clarification — cards are always expected to be 10 USDe.

Checks: reviewed `c5390d880430..aac0ccdb5d6c`; `git diff --check origin/bingo...HEAD` passed. Local Vitest still could not run because dependencies are unavailable in this sandbox (`vitest: not found` after inode-exhausted install attempts). Note: PR is still marked draft.

PR #1878, opened 2026-06-17
feat(api): v2 Account.statsHistory (+ VaultStat parity) and ConditionGroup groupIds

noahlitvin:
## What & why

The exchange-ui **Portfolio → Prediction Stats** page reads four per-account time-series that only existed on the v1 API (PnL chart + calendar, volume, win-rate counts, pending claims + balance). v2 had dropped them (lifetime aggregates only). This adds them back as **`Account.statsHistory`** so exchange-ui can move fully onto `/v2/graphql`, and aligns the new surface with the existing `VaultStat`.

Also adds a small **`ConditionGroup.groupIds`** filter for the markets-by-id hydration path.

## `Account.statsHistory`

New bucketed series on `Account` — `AccountStatPoint` per `interval` (HOUR/DAY/WEEK/MONTH), oldest-first, default window with `first`/`after` paging. Computed live (no snapshot table) from the shared `services/timeSeriesQueries` SQL — the same source the v1 `accountBalance`/`accountPnl`/`accountVolume`/`accountPredictionCount` queries read — so it needs no new writer. Fields: `deployedCollateral`, `claimableCollateral`, `volume`, `realizedPnl`, `cumulativePnl`, and prediction counts (total/won/lost/pending/non-decisive).

## VaultStat / AccountStat consistency

Kept as **separate types** (vault = daily snapshot, chain-scoped, superset with AUM/flows; account = live, cross-chain), with consistency at the vocabulary level plus the existing `Vault.account` edge:

- **Renamed `VaultStat.unredeemedClaim` → `claimableCollateral`** across the monorepo (SDL, resolver, SDK query + types, app TVL adapter) to match `AccountStatPoint.claimableCollateral`. The internal DB column (`vaultUnredeemedClaim`) keeps its name.
- **`Vault.statsHistory` now returns oldest-first (ascending)** to match `Account.statsHistory`. The SDK/app already re-sorted ascending, so this is transparent to the dashboard.

### PnL realizes at redemption (both account and vault)

`cumulativePnl` realizes PnL **only when a position is actually settled/redeemed** (a `Claim`/`Close`), plus secondary-market cash flow — it is **not** marked from still-unredeemed `claimableCollateral`.

Why: a win can change hands on the secondary market before redemption. Marking PnL from unredeemed claimable attributes the win by the *original minter*, not the current token holder — so after someone sells a winning token pre-redeem, the seller keeps showing the win and the buyer shows none, until redemption. A redemption records the **true holder**, so realizing only then is correct by construction and needs no holder-balance reconstruction: the seller keeps just their sale proceeds, the buyer realizes the win on redeeming.

Both types use this same rule, so they stay consistent. Trade-off (documented in the SDL + resolver comments): a resolved-but-unredeemed win isn't marked until claimed — the line steps up at redemption rather than at resolution. `claimableCollateral` is still surfaced on both (pending-claims view / vault TVL line), just never folded into PnL.

## PnL correctness fixes (shared `timeSeriesQueries` helper)

- **Running `cumulativePnl`:** seeds the line from PnL realized *before* the window instead of resetting to 0 at the window start; per-bucket `realizedPnl` stays window-local.
- **Secondary-market PnL:** `queryAccountPnl` books secondary trades (`sold − bought`), mirroring the vault's convention. Volume already counted secondary; PnL didn't.
- **Per-interval default window in `resolveDefaults`:** an unfiltered `HOUR` request returns the last 7 days instead of throwing `"Too many buckets"` (the 90-day default = 2160 hourly buckets). DAY/WEEK/MONTH unchanged. Shared helper, so it also fixes the equivalent v1 queries (bug fix, no in-repo v1 consumers).

## Tests

- Real-DB integration tests for running cumulative + secondary-market PnL (`…pnl.integration`) and claimable accounting (`…balance.integration`).
- Unit tests for connection shaping, the realize-at-redemption rule (a large `claimableCollateral` with no realized PnL must not move the line), the `resolveDefaults` HOUR clamp, and the `groupIds` filter.

## Notes for reviewers

- The `claimableCollateral` rename and ascending-order change touch the **SDK and app**, not just the API — intentional, to keep consumers in sync.
- Generated artifacts (`schema.v2.graphql`, SDK `graphql.v2.ts`) are regenerated and in sync.
- Residual second-order difference: in the rare "sell a winning token pre-redeem" case, the vault books stake at *resolution* while the account books it at *claim*, so the seller's stake is treated slightly differently. Fully closing that means moving the vault's stake timing to claim (snapshot-aggregator change) — out of scope. The user-facing issue (crediting a sold-off win to the wrong party) is resolved on both.

### DIFF
--- a/packages/api/schema.v2.graphql
+++ b/packages/api/schema.v2.graphql
@@ -180,6 +180,12 @@ type AccountStatPoint {
   """
   Settled-and-won but not-yet-redeemed collateral owed to the account as of
   the bucket — the account analog of `VaultStat.claimableCollateral`.
+
+  This is a *pending-balance* figure, deliberately NOT part of `cumulativePnl`
+  (see there). It is attributed to the original position side and does not
+  follow secondary-market transfers, so after a winning position token is sold
+  pre-redemption it still reads against the seller until redeemed. Safe for the
+  "pending claims" view; not used to realize PnL precisely for that reason.
   """
   claimableCollateral: BigInt!
 
@@ -190,21 +196,31 @@ type AccountStatPoint {
   volume: BigInt!
 
   """
-  PnL realized within the bucket: settlement gains − losses on positions that
-  settled in-window, plus secondary-market sale proceeds net of purchase cost
-  (`sold − bought`) on trades executed in-window. Same trading-PnL definition
-  the vault plots in `VaultStat.cumulativePnl`, measured per-bucket.
+  PnL realized within the bucket — the per-bucket increment of `cumulativePnl`
+  (see there for exactly when PnL realizes): redemption gains on wins claimed
+  in-window, losses/non-decisive at settlement, and secondary-market proceeds
+  net of cost (`sold − bought`) on trades executed in-window. Same trading-PnL
+  definition the vault plots, measured per-bucket.
   """
   realizedPnl: BigInt!
 
   """
   Running trading PnL through the end of the bucket: the cumulative sum of
-  `realizedPnl` (seeded by all PnL realized *before* the window, so the line
-  never resets at the window start), plus the net winnings on positions that
-  have settled in the account's favour but are not yet redeemed
-  (`claimableCollateral` minus the stake on those wins). That unredeemed mark
-  makes a resolved win show up at resolution rather than waiting for redemption
-  — the account analog of `VaultStat.cumulativePnl`.
+  `realizedPnl`, seeded by all PnL realized *before* the window so the line
+  never resets at the window start.
+
+  PnL is realized only when a position is actually settled — a win at the
+  moment it is *redeemed* (a claim), a loss/non-decisive at settlement, plus
+  secondary-market proceeds and cost as they trade. It is NOT marked from
+  still-unredeemed `claimableCollateral`. This is deliberate: a redemption
+  records the true holder, so a win is credited to whoever actually collects
+  it. If a winning position token is sold before redemption, the seller simply
+  keeps their sale proceeds and the buyer realizes the win on redeeming — no
+  prize is ever credited to someone who no longer holds the token. The cost is
+  cosmetic: a resolved-but-unredeemed win contributes nothing until it is
+  claimed (the line steps up at redemption rather than at resolution).
+
+  Matches `VaultStat.cumulativePnl`, which realizes the same way.
   """
   cumulativePnl: BigInt!
 
@@ -365,14 +381,17 @@ type VaultStat {
   realizedPnl: BigInt!
 
   """
-  Cumulative trading PnL the vault dashboard plots: settlement PnL, plus
-  wUSDe earmarked from resolved-but-unredeemed wins, plus net secondary-
-  market flow (`realizedPnl + claimableCollateral + secondarySold −
-  secondaryBought`). This is the *same* numerator the account behind the
-  vault carries; the vault presents it as a return on total assets
-  (`balance + deployedCollateral`), where the account measures it against
-  deployed capital. The unredeemed term cancels the transient phantom loss
-  between a position resolving and its redeem being indexed.
+  Cumulative trading PnL the vault dashboard plots: settlement PnL plus net
+  secondary-market flow (`realizedPnl + secondarySold − secondaryBought`).
+
+  PnL realizes only when a position is actually redeemed/settled — it is NOT
+  marked from still-unredeemed `claimableCollateral` (that term used to be
+  added here and has been removed). This keeps the line correct when a winning
+  token is sold before redemption: the redeemer collects the prize, so it is
+  never credited to whoever merely minted the position. Realizes the same way
+  as `AccountStatPoint.cumulativePnl`. Trade-off: a resolved-but-unredeemed win
+  is not marked until it is claimed — the line steps up at redemption rather
+  than at resolution, which can read as a brief paper dip in between.
   """
   cumulativePnl: BigInt!
   deposits: BigInt!
@@ -386,8 +405,10 @@ type VaultStat {
   wUSDe owed to the vault on resolved-but-not-yet-redeemed winning sides,
   net of what it has already claimed. The dashboard's TVL line adds this to
   `balance + deployedCollateral` so funds that have settled in the vault's
-  favor but await an indexed redeem are not transiently dropped. Named to
-  match `AccountStatPoint.claimableCollateral`, the account-level analog.
+  favor but await an indexed redeem are not transiently dropped. A
+  pending-balance figure for TVL only — deliberately NOT part of
+  `cumulativePnl` (PnL realizes at redemption). Named to match
+  `AccountStatPoint.claimableCollateral`, the account-level analog.
   """
   claimableCollateral: BigInt!
 }
--- a/packages/api/schema.v2.graphql
+++ b/packages/api/schema.v2.graphql
@@ -198,10 +198,13 @@ type AccountStatPoint {
   realizedPnl: BigInt!
 
   """
-  Running trading PnL through the end of the bucket — the cumulative sum of
-  `realizedPnl`, seeded by all PnL realized *before* the window so the line
-  never resets at the window start. The account analog of
-  `VaultStat.cumulativePnl`.
+  Running trading PnL through the end of the bucket: the cumulative sum of
+  `realizedPnl` (seeded by all PnL realized *before* the window, so the line
+  never resets at the window start), plus the net winnings on positions that
+  have settled in the account's favour but are not yet redeemed
+  (`claimableCollateral` minus the stake on those wins). That unredeemed mark
+  makes a resolved win show up at resolution rather than waiting for redemption
+  — the account analog of `VaultStat.cumulativePnl`.
   """
   cumulativePnl: BigInt!
 
--- a/packages/api/schema.v2.graphql
+++ b/packages/api/schema.v2.graphql
@@ -220,7 +220,17 @@ type AccountStatPoint {
   cosmetic: a resolved-but-unredeemed win contributes nothing until it is
   claimed (the line steps up at redemption rather than at resolution).
 
+  Secondary sales also book the seller's allocated mint cost basis
+  (`stake × tokensSold / minted`), so a seller who never redeems still has
+  their stake deducted and combined PnL across buyer + seller conserves.
+
   Matches `VaultStat.cumulativePnl`, which realizes the same way.
+
+  Known approximations (tracked for a planned PnL rework to a single cash-flow +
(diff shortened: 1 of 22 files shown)


nostradamus-agent (changes requested):
🔮 Review verdict: **request changes**.

Found three blockers:

1. **Generated v2 SDK types are stale / missing.**
   `packages/api/codegen.ts:34-36` explicitly generates `packages/sdk/types/graphql.v2.ts` from `packages/api/schema.v2.graphql`, but the PR updates the v2 schema without committing the regenerated SDK type output. I ran:
   `DATABASE_URL='postgresql://user:***@localhost:5432/db' pnpm --filter @sapience/api run generate-types`
   and it produced a real diff in `packages/sdk/types/graphql.v2.ts` (+106 lines: `Account.statsHistory`, `AccountStatPoint*`, `AccountStatFilter`, `ConditionGroupFilter.groupIds`, `TimeInterval`). This will leave v2 SDK consumers typed against the old schema.

2. **`cumulativePnl` is window-local, not cumulative as the SDL promises.**
   The v2 field docs say `cumulativePnl` is “running settlement PnL through the end of the bucket” and mirrors `VaultStat.cumulativePnl` (`schema.graphql:238-242`). But `Account.statsHistory` forwards the requested/default `from` into `queryAccountPnl`, and that SQL filters every PnL event to `fromEpoch..toEpoch` before doing `SUM(...) OVER (ORDER BY bucket_epoch)` (`timeSeriesQueries.ts:260-302`). So the default 90-day response resets cumulative PnL at the window start and drops all prior realized PnL.

3. **Secondary-market PnL is omitted while the new API claims it is included.**
   The new resolver docs and SDL say this history is computed from “claims, closes, and secondary trades” (`accountStatsHistory.ts:8-9`, `schema.graphql:162-164`). Volume includes `secondary_trade`, but `queryAccountPnl` only unions Claim, Close, and legacy settled `position`; there is no secondary-trade PnL leg. That makes buyer cost basis and seller proceeds wrong for any account with secondary trades.

Verification performed:
- `pnpm install --frozen-lockfile --prefer-offline`
- Targeted v2 tests pass: `accountStatsHistory.test.ts`, `AccountStatsHistory.test.ts`, `conditionGroup.test.ts` — 15/15
- `pnpm --filter @sapience/api run type-check` passes
- `pnpm --filter @sapience/api run generate-types` exposes the stale SDK generated-types diff above

Non-blocking: `filter: { groupIds: [] }` currently means “no groupIds filter” because the resolver gates on `.length`; if exchange-ui might pass an empty hydration list through, consider returning an empty connection instead of the default page. Not blocking if the caller short-circuits empty lists.

nostradamus-agent (changes requested):
Re-reviewed the new commits. Better — the stale SDK artifact is fixed and the cumulative baseline no longer resets at the requested window start. Still requesting changes for two API/correctness blockers:

1. `VaultStat.unredeemedClaim` was removed/renamed to `claimableCollateral`. That field exists on `origin/staging` (`VaultStat.unredeemedClaim: BigInt!`), so existing v2 clients validating a query with `unredeemedClaim` will hard-fail after this deploy. Please keep it as a deprecated alias alongside `claimableCollateral` unless we’re intentionally doing a breaking v2 schema cut. Evidence: `packages/api/src/graphql/v2/schema/schema.graphql:431-438`, `packages/api/src/graphql/v2/resolvers/Vault.ts:135-140`, `packages/sdk/queries/vault.ts:47-49`.

2. `Account.statsHistory` claimable/unredeemed PnL is still original-position-side based, not actual-token-holder based after secondary sales. `queryAccountBalance` builds `all_claimable` only from `Prediction.predictor/counterparty` (`timeSeriesQueries.ts:405-425`) and only stops counting after that same address has a matching `Claim` (`:431-460`). If A sells a winning position token to B before redemption, A keeps showing `claimableCollateral` / `unredeemedNetGain` / inflated `cumulativePnl` until B redeems; B shows no pre-redeem claimable mark. That contradicts the SDL’s “collateral owed to the account” semantics and the PR’s secondary-aware stats goal. Please cover the secondary transfer case in tests.

Non-blocking footgun: `filter.timestamp.gte` is documented as an inclusive timestamp window, but the live grid starts at `date_trunc(interval, gte)`, so a noon daily lower bound returns a midnight bucket and can include midnight→noon bucket PnL. If callers only send aligned bounds, fine; otherwise document/normalize/reject.

Verification I ran locally:
- API targeted tests: 43/43 passed (`accountStatsHistory`, `AccountStatsHistory`, `conditionGroup`, `timeSeriesQueries`, `Vault`).
- SDK vault query tests: 7/7 passed.
- App vault adapter/chart tests: 8/8 passed.
- `@sapience/api` `type-check`: passed.
- `@sapience/api` `generate-types`: passed, no diff — generated v2 SDK/schema artifacts are fresh now.
- `@sapience/sdk` build and `@sapience/app` type-check: passed.

🔮 Much closer, but the secondary holder accounting bug will make the shiny new account chart lie exactly when users trade the winning token. That’s the worst time to lie.

nostradamus-agent (changes requested):
Re-reviewed at `63a8d102`.

No longer blocking on the `VaultStat.unredeemedClaim` → `claimableCollateral` rename given Noah’s note that v2 consumers are only in this repo. I verified the repo callers are migrated and the v2 generated artifacts are fresh.

Still requesting changes for one accounting bug in the latest PnL semantics:

`queryAccountPnl` now fixes the previous “credit unredeemed wins to the original minter” issue by realizing wins only at redemption. Good direction. But secondary-sale PnL still books seller proceeds as pure `+price` with no minted-cost basis deduction (`packages/api/src/services/timeSeriesQueries.ts:316-325`). The only place original stake basis is deducted is the Claim branch (`:258-266`), and that only runs for the address that redeems. So if A mints/stakes 20 on the eventual winning token, sells those tokens to B for 60 before redemption, and B redeems 100:

- A currently reports `+60` from the sale, not `+40` (`60 - 20` basis).
- B reports `-60 + 100 = +40`.
- Combined PnL becomes `+100`, but the real combined profit is `+80` (the loser’s stake). The original minter’s stake disappeared from PnL unless that same minter redeems.

The new integration tests cover claim basis allocation and generic buy/sell cash flow, but not this exact “original minter sells winning tokens before buyer redeems” path (`packages/api/src/services/timeSeriesQueries.pnl.integration.test.ts:135-142`, `:223-238`). Please add that case and allocate original mint basis on secondary sales too — likely by joining `secondary_trade.token`/`tokenAmount` through `Picks` into the same `claim_stake` basis and subtracting `stake * tokenAmount / minted` for seller-side trades when the seller is the original side holder. That also preserves basis when they sell some tokens and redeem the rest.

Verification I ran:
- API targeted v2/service tests: 43/43 passed.
- SDK vault query tests: 7/7 passed.
- App vault adapter/chart tests: 8/8 passed.
- `@sapience/api` type-check: passed.
- `@sapience/api` generate-types: passed, no diff.
- `@sapience/sdk` build: passed.
- `@sapience/app` type-check: passed.
- Worktree clean.

🔮 Very close. The last bug is just the basis falling through the floor when the original winner sells before the buyer redeems.

nostradamus-agent (changes requested):
Re-reviewed at `be4047d06`.

The specific secondary-sale bug from the last review is fixed for the covered path: original minter sells all winning tokens, buyer redeems, seller’s mint basis is deducted on sale. The new regression test captures that case.

Still requesting changes for a different account-stats correctness hole that this PR now exposes through `Account.statsHistory.claimableCollateral`:

`queryAccountBalance` removes an entire `(pickConfigId, side)` from claimable collateral as soon as *any* matching claim exists (`packages/api/src/services/timeSeriesQueries.ts:471-483`). It only checks `NOT EXISTS account_claims`, keyed by `pickConfigId + side`, and ignores `Claim.tokensBurned` / `Claim.collateralPaid`.

That is wrong for partial redemption:

- A has a settled winning side with `predictorClaimable = 100`.
- A redeems 40 tokens / 40 collateral.
- Remaining claimable should still be 60 until the rest is redeemed.
- Current query reports 0 from the first claim onward because the `NOT EXISTS` fails for the whole side.

This matters more now because the PR adds `Account.statsHistory.claimableCollateral` as a first-class v2 field and describes it as pending claimable balance. The same file already treats partial claims as real in the PnL path (`tokensBurned` proportional basis), and the integration tests explicitly cover partial exits, so the balance side should not assume “one claim means the whole side is gone.”

Likely fix: aggregate claims by `(pickConfigId, side, redeemedAt)` and subtract claimed collateral/token-proportional claimable from the outstanding amount per bucket, rather than excluding the whole side on first claim. Add a regression test with a partial claim where claimable drops from 100 → 60 → 0 across buckets.

Non-blocking but still worth tracking: the sale-basis patch still has the documented mixed-inventory approximation — if an address mints, buys the same token, sells/redeems more than its minted inventory, minted basis can be allocated to purchased tokens too. You already call this out in the SDL, so I’m not using it as the blocker here, but the oracle remains grumpy about it.

Verification I ran:
- API targeted-ish test command passed; due the repo’s vitest arg wiring it ran the API suite: 826 passed / 10 skipped.
- `@sapience/api` type-check passed.
- `@sapience/api` generate-types passed; worktree stayed clean.
- `@sapience/sdk` tests passed; same arg wiring ran the SDK suite.
- `@sapience/sdk` build passed.
- `@sapience/app` tests passed; same arg wiring ran the app suite.
- `@sapience/app` type-check passed.
- Direct `timeSeriesQueries.pnl.integration.test.ts` run skipped locally because no Postgres test DB is reachable in this sandbox.

🔮 Close, but this one would make partial claims disappear from the pending-claimable chart. That’s the kind of tiny accounting ghost that becomes a support thread.

nostradamus-agent (approved):
Re-reviewed the new `e1b61cd` fix. Ship it.

What changed since my last pass: `queryAccountBalance` now computes claimable as gross owed per `(pickConfigId, side)` minus summed `Claim.collateralPaid` through the bucket, floored at zero. That fixes the partial-redemption case without regressing the redeemed-side matching fix.

Verified:
- `pnpm --filter @sapience/api exec vitest run src/services/timeSeriesQueries.balance.integration.test.ts src/services/timeSeriesQueries.test.ts src/services/timeSeriesQueries.pnl.integration.test.ts` → 19 passed, DB integration tests skipped locally because no Postgres.
- `pnpm --filter @sapience/api run lint` → passed.
- `pnpm --filter @sapience/api run type-check` → passed.

Non-blocker: the new integration coverage is still only useful in environments with `TEST_DATABASE_URL`/local Postgres; locally it skips cleanly. But the SQL shape and added partial-redemption case look right.

PR #1882, opened 2026-06-18
feat: sort positions by resolution time + robust resolvedAt

noahlitvin:
## What

Lets users sort their `/profile` positions by **when each position was considered resolved on Sapience**, and makes that timestamp trustworthy.

## Why

`Picks.resolvedAt` previously borrowed the *triggering* condition's `settledAt`. For combos that's wrong when conditions settle out of order, and a config minted *after* all its conditions already settled would never resolve at all (no future settlement event fires for it).

## Changes

**Backend — data correctness** (`resolvePickConfigs.ts`, `predictionMarketEscrowIndexer.ts`)
- `resolvedAt` now derives from leg data:
  - **Loss** (`COUNTERPARTY_WINS`) → earliest adverse leg's `settledAt` (the leg that doomed the combo).
  - **Win** (`PREDICTOR_WINS`) → latest leg's `settledAt` (last leg to settle).
  - Falls back to the trigger timestamp only when leg-level `settledAt` is missing, so existing behavior is unchanged when leg data is present.
- New `resolvePickConfig()` runs at mint (`PredictionCreated`): settles a config from current condition state when its conditions already settled before it was created — fixes the latent never-resolves case.

**Frontend — `/profile` positions table** (`PositionsTable.tsx`)
- New sortable **"Resolved"** column showing relative resolution time, with a muted `PENDING` for unresolved positions (matches the Profit/Loss column).
- Client-side sort consistent with the table's existing columns; default DESC (most recently resolved first), pending sinks to the bottom.

## Notes

- **No schema change / migration.** `resolvedAt` was already exposed (`PickConfiguration.resolvedAt`) and already fetched by the FE — it just needed to be correct.
- Server-side `PositionOrderField` sort intentionally not added: the table sorts client-side (as do all its other columns), so a keyset `orderBy` on a nullable related Int would be unused complexity.

## Testing

- API: full suite (847 tests) green; added unit coverage for the first-loss/last-win timestamp logic and all four resolve-at-mint cases.
- `tsc` clean (api + app); ESLint clean on changed files.
- Verified end-to-end against the live DB locally: `positions` returns populated `resolvedAt` on settled configs and `null` on unresolved.

### DIFF
--- a/packages/api/src/graphql/v2/resolvers/queries/position.ts
+++ b/packages/api/src/graphql/v2/resolvers/queries/position.ts
@@ -34,6 +34,7 @@ import {
   encodeCursor,
   normalizeDirection,
   withCursorWhere,
+  type OrderDirection,
 } from '../../relay/connection';
 import { tryFromGlobalIdV2 } from '../../relay/nodeRegistry';
 
@@ -59,9 +60,19 @@ export const position: NonNullable<QueryResolvers['position']> = async (
   });
 };
 
-const FIELD_TO_PRISMA: Record<string, 'createdAt' | 'updatedAt'> = {
+/**
+ * CREATED_AT / UPDATED_AT keyset on a Date column of the Position row
+ * itself. RESOLVED_AT orders by the related `pickConfiguration.resolvedAt`
+ * (Unix seconds, nullable) — handled on a separate branch in `positions`
+ * because it keysets through the relation and restricts to resolved rows.
+ */
+const FIELD_TO_PRISMA: Record<
+  string,
+  'createdAt' | 'updatedAt' | 'resolvedAt'
+> = {
   CREATED_AT: 'createdAt',
   UPDATED_AT: 'updatedAt',
+  RESOLVED_AT: 'resolvedAt',
 };
 
 /** Raw Position row as fetched for synthesis (predictions included). */
@@ -93,7 +104,17 @@ export type WacFields = {
     | null;
   /** Trade hash of the sale this SOLD row represents; null on OPEN rows. */
   soldTradeHash: string | null;
-  cursorRow: { id: number; createdAt: Date; updatedAt: Date };
+  /**
+   * Order/keyset values pinned to the underlying RAW row. `resolvedAt`
+   * (from the pickConfiguration, Unix seconds) travels alongside the Date
+   * columns so the RESOLVED_AT keyset cursor can read it directly.
+   */
+  cursorRow: {
+    id: number;
+    createdAt: Date;
+    updatedAt: Date;
+    resolvedAt: number | null;
+  };
 };
 
 export type SynthesizedPositionRow = RawPositionRow & WacFields;
@@ -265,7 +286,12 @@ const synthesizeWacRows = async (
           status: 'SOLD',
           prediction,
           soldTradeHash: d.tradeHash,
-          cursorRow: r,
+          cursorRow: {
+            id: r.id,
+            createdAt: r.createdAt,
+            updatedAt: r.updatedAt,
+            resolvedAt: r.pickConfiguration?.resolvedAt ?? null,
+          },
         });
       }
     }
@@ -285,7 +311,12 @@ const synthesizeWacRows = async (
         status: 'OPEN',
         prediction,
         soldTradeHash: null,
-        cursorRow: r,
+        cursorRow: {
+          id: r.id,
+          createdAt: r.createdAt,
+          updatedAt: r.updatedAt,
+          resolvedAt: r.pickConfiguration?.resolvedAt ?? null,
+        },
       });
     }
   }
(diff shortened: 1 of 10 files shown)


packages/app/src/components/positions/PositionsTable.tsx
```diff
@@ -398,6 +398,20 @@ function PositionRow({
           );
         })()}
       </TableCell>
+      {/* Resolved — when Sapience recorded the on-chain resolution */}
+      <TableCell className="whitespace-nowrap">
+        {pickConfig?.resolvedAt ? (
```
nostradamus-agent, line 407: This treats `resolvedAt: null` as pending, but `resolvedAt` is nullable independently of `pickConfig.resolved` (`resolvedAt Int?`), and the backend can still produce `resolved: true, resolvedAt: null` for legacy/missing-leg-timestamp paths. The row can show `WON`/`LOST` in the status column and `PENDING` here. Gate pending on `!pickConfig?.resolved`; for resolved-without-timestamp render an unknown/blank state instead.

packages/app/src/components/positions/PositionsTable.tsx
```diff
@@ -671,6 +694,10 @@ export default function PositionsTable({
           // Missing end time sinks to the bottom regardless of direction
           return endsAt === 0 ? Number.POSITIVE_INFINITY : endsAt;
         }
+        case 'resolvedAt':
+          // When Sapience recorded the resolution. Still-pending positions
+          // (null) sort to 0 → bottom of the default DESC (recent-first) view.
+          return p.pickConfig?.resolvedAt ?? 0;
```
nostradamus-agent: This sorts only `filteredPositions`, i.e. the cursor pages already loaded by `usePositionBalances` (default 15 rows). So the new Resolved sort is not global: a newer resolved position on page 2 will not bubble above page-1 rows until the user loads more. For a `Resolved` sort column, the ordering needs to be pushed into the positions query/API, or the UI needs to make the loaded-only limitation explicit.

nostradamus-agent (changes requested):
Verdict: request changes.

Two correctness issues in the new Resolved sort/display:

- Resolved positions with `resolvedAt = null` render as `PENDING`, even though `resolved` can be true while the timestamp is unknown/legacy/missing. That makes won/lost rows look pending.
- The new sort is only over currently loaded cursor pages. With page size 15, clicking Resolved does not globally sort a user's positions by resolution time; newer/later rows can stay hidden behind pagination.

Backend resolution logic looks sane: loss = earliest adverse leg, win = latest leg, and mint-time resolution covers already-settled configs. Could not run full tests locally because this sandbox is out of inodes during `pnpm install`; `git diff --check` passed.

packages/api/src/graphql/v2/resolvers/queries/position.ts
```diff
@@ -337,18 +400,27 @@ export const positions: NonNullable<QueryResolvers['positions']> = async (
     pickConfigFilter.endsAt = r;
     hasPickConfigFilter = true;
   }
+  // RESOLVED_AT ordering only makes sense for resolved positions; pinning
+  // resolvedAt non-null also keeps the keyset clean (no NULL boundary).
+  if (byResolvedAt) {
+    pickConfigFilter.resolvedAt = { not: null };
```
nostradamus-agent: This makes `orderBy: RESOLVED_AT` a filter as well as a sort. The UI column still presents this as sorting the positions table, with pending positions supposed to sink to the bottom, but the API now excludes every unresolved row (`resolvedAt: null`). Worse, if the user has the Active status filter selected, the request combines `resolved=false` with `resolvedAt != null`, so sorting by Resolved can return an empty table. Sorting should not silently narrow the result set; either include nulls with a deterministic nulls-last/nulls-first strategy, or make the UI explicitly switch to a resolved-only view.

packages/app/src/components/positions/PositionsTable.tsx
```diff
@@ -482,6 +511,18 @@ export default function PositionsTable({
   const [filters, setFilters] = React.useState<PositionsFilterState>(
     getDefaultPositionsFilterState
   );
+  const [sort, setSort] = React.useState<SortState | null>(null);
+
+  // RESOLVED_AT sorts server-side so it orders across the whole resolved set
```
nostradamus-agent, line 516: This says RESOLVED_AT orders across the whole resolved set, but the table header/action reads as a normal sort of the current positions set. Because the backend drops unresolved rows for this order, users lose active/pending positions when they click the column instead of seeing them sink below resolved rows. That conflicts with the existing `PENDING` rendering and the previous PR description.

nostradamus-agent (changes requested):
Verdict: still request changes.

The earlier `resolvedAt: null` display issue is fixed, and the backend now exposes a RESOLVED_AT order path. But the fix changes table semantics: clicking Resolved now filters the connection to `resolvedAt != null`, so active/pending positions disappear instead of sinking to the bottom. With the Active status filter, the query becomes `settled=false AND resolvedAt != null`, so the table can go empty just by sorting.

Verification: reviewed delta f6faa47..cc10577 and ran `git diff --check`. Could not run package tests/typecheck here because the sandbox is still inode-exhausted for pnpm install.

nostradamus-agent (approved):
Verdict: approved.

The previous blocker is fixed: RESOLVED_AT now keeps null/pending rows and orders them NULLS LAST instead of filtering them out, with cursor handling for both non-null and null phases. Active filter + Resolved sort no longer collapses to `resolved=false AND resolvedAt != null`.

Checks: reviewed delta cc10577..0c4828e and ran `git diff --check` successfully.

Non-blocking nit: the GraphQL/schema/SDK comments still say RESOLVED_AT restricts to resolved/non-null rows; worth updating so the docs match the fixed behavior. I’m not blocking on that.

PR #1886, opened 2026-06-19
perf(api): speed up /analytics — daily-bucket Protocol.statsHistory + memoize TVL/cross-family reads

noahlitvin:
## Problem

The `/analytics` page was slow because the frontend loaded the protocol snapshot series by **paginating it 100 rows at a time, sequentially, up to a 50-page cap**. Against the production cadence (snapshots every ~4 hours → ~5000 rows) that meant:

- **~88s of wall-clock** across 50 sequential `v2/graphql` requests (measured against the read-only prod DB).
- The series is **oldest-first**, so hitting the 50-page cap **silently truncated the most recent history** — the charts were quietly missing recent days.
- Each `statsHistory` page re-ran two uncached reads (a full `protocolStatsSnapshot` scan + the per-vault TVL fan-out), and `Protocol.stats` re-ran the TVL fan-out on every request.

The charts only ever render this data **bucketed to daily points**, so the fine-grained 4-hourly series was fetched only to be thrown away client-side.

## Fix (A + B + C)

**A — server-side daily downsampling.** Add `interval: TimeInterval` to `Protocol.statsHistory`. When set, the resolver collapses the snapshot series into fixed-width buckets: stock fields (`cumulative*`, `openInterest`, `escrowBalance`, `totalValueLocked`) take the bucket's last/latest row; flow fields (`period*`) sum across the bucket; the node `timestamp` is the bucket start (matching the client's existing `floor(ts/86400)*86400` grid, so the client's own bucketing becomes an idempotent no-op — **no UI component changes needed**). The SDK now requests `interval: DAY` with `first` omitted, so the full ~250-point daily series returns in **one request**. With no `interval` the raw series is returned, so the change is **backward compatible**.

**B — memoize `crossFamilyAvailableByRawTs`** (60s single-flight per chainId) so `statsHistory` stops re-reading the full snapshot table on every page/request.

**C — memoize `computeProtocolTvl`** likewise, so `Protocol.stats` stops re-running the per-vault latest-snapshot fan-out on every request.

## Result (measured vs read-only prod DB)

| | Before | After |
|---|---|---|
| Analytics history load | ~88s (50 sequential requests) | **1 request, ~8s cold / 6ms warm** |
| Recent-day truncation | yes (50-page cap) | **no** (251 daily nodes, `hasNextPage: false`) |

## Schema change

Additive and non-breaking — one new optional argument, reusing the existing `TimeInterval` enum (already used by `Account.statsHistory`):

```diff
- statsHistory(first: Int = 100, after: String, filter: ProtocolStatFilter): ProtocolStatConnection!
+ statsHistory(interval: TimeInterval, first: Int = 100, after: String, filter: ProtocolStatFilter): ProtocolStatConnection!
```

## Tests

- `Protocol.test.ts`: new test for `interval: DAY` bucketing (last-of-day stock fields, summed flows, bucket-start label, cross-family TVL), plus memoization tests for B and C. Existing raw-path / filter / live-candle tests still pass.
- `protocol.test.ts` (SDK): updated for the new query shape (`interval: DAY`, `first: null`) and single-page fetch.
- Type-check + lint clean; verified end-to-end against the live API.

### DIFF
--- a/packages/api/schema.v2.graphql
+++ b/packages/api/schema.v2.graphql
@@ -1690,7 +1690,7 @@ type Protocol {
   Live, current protocol-wide stats — a single `ProtocolStat` with
   `timestamp = now`. Volume / trade-count / open-interest reflect the
   in-progress period and `totalValueLocked` is read across every configured
-  vault at query time. Use `statsHistory` for the recorded daily series.
+  vault at query time. Use `statsHistory` for the recorded snapshot series.
   """
   stats: ProtocolStat!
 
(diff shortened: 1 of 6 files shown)


nostradamus-agent (approved):
🔮 Reviewed — ship it.

I traced the v2 `Protocol.statsHistory(interval:)` path, the daily bucketing semantics, the SDK query change to `interval: DAY`/`first: null`, and the cache single-flight behavior. No blocking issues found.

Verified locally:
- `packages/api/node_modules/.bin/vitest run packages/api/src/graphql/v2/resolvers/Protocol.test.ts` — 6 passed
- `packages/sdk/node_modules/.bin/vitest run packages/sdk/queries/__tests__/protocol.test.ts` — 11 passed
- `prettier --check` on changed TS files — passed
- schema prettier drift exists on base too, so not introduced by this PR
- diff secret scan — clean

GitHub status visible here: Vercel app/docs green. API-specific CI isn’t showing up in checks, hence the local focused tests above.

PR #1889, opened 2026-06-19
Scope position related activity by token

nostradamus-agent:
## Summary
- adds a v2 `ActivityFilter.token` field for exact position-token scoped activity
- scopes prediction activity side-aware when `account + token` are provided
- scopes secondary trades to the exact token and passes `position.tokenAddress` from the position dialog

## Why
Related Activity in the position dialog should describe the token balance, not every activity row for the same account/pick config. This keeps older valid trades for the same token, while excluding unrelated secondary trades.

## Tests
- `pnpm --filter @sapience/api run type-check`
- `pnpm --filter @sapience/app run type-check`
- `pnpm --filter @sapience/api exec vitest run src/graphql/v2/resolvers/queries/activity.test.ts`
- `pnpm --filter @sapience/app exec vitest run src/hooks/graphql/__tests__/useAccountActivity.test.ts src/components/positions/__tests__/PositionDialog.test.tsx`
- `pnpm --filter @sapience/api run lint`
- `pnpm --filter @sapience/app run lint`

PR #1885, opened 2026-06-18
Staging -> Main

ukitta555:
Promotes `staging` to `main`.

### Included (diff vs `main`)

**fix(market-keeper): migrate relist discovery to keyset pagination (#1884)**

The `discovery` cron was crashing in `fetchPastEndDateMarkets`: it paged Polymarket's Gamma `/markets` endpoint by offset, and once the past-endDate corpus crossed ~2000 markets the API rejected `offset=2100` with `HTTP 422 ("offset too large, use /markets/keyset for deeper pagination")`. 422 is non-retryable, so the whole `cron-discovery` group exited 1 every run.

- Migrated to the `/markets/keyset` endpoint: reads the `{ markets, next_cursor }` envelope and advances with `after_cursor` (no offset → no 2000 cap → no 422).
- Terminates on absent `next_cursor`; guards against a stuck cursor (`seenCursors`) + `MAX_PAGES` backstop.
- Logs the earliest `endDate` per page for visible forward progress.
- New `relist-market.test.ts` covers pagination, termination, and the stuck-cursor guard.

Verified live: walks 28 pages / 2664 markets in ~6.6s with no 422.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1891, opened 2026-06-19
Staging → Main

noahlitvin:
## Summary

Promotes the current `staging` branch to `main`. Net diff spans the API, app (frontend), and SDK packages. The substantive changes (by squash-merged PR):

- **#1884 — fix(market-keeper): migrate relist discovery to keyset pagination.** Replaces offset pagination in relist discovery with keyset pagination.
- **#1886 — perf(api): speed up `/analytics`.** Daily-bucket `Protocol.statsHistory` + memoize TVL / cross-family reads (downsampling).
- **#1888 — Add Singles Vault to vault tabs.**
- **#1889 — Scope position related activity by token.** Activity in the position dialog is now scoped by token alone.
- **#1890 — fix: use indexed vault account value for vaults AUM.** Vaults AUM uses the indexed vault account value; PnL chart loader tracks the vault-stats query rather than account value.

## Diff overview

```
20 files changed, 807 insertions(+), 106 deletions(-)
```

Touched areas: `packages/api` (Protocol/Activity resolvers + schema), `packages/app` (vaults page, position dialog, activity table, analytics hook), `packages/sdk` (protocol/vault queries + generated v2 types).

## Review

A thorough review of this diff follows in the PR comments.

### DIFF
--- a/packages/api/schema.v2.graphql
+++ b/packages/api/schema.v2.graphql
@@ -359,8 +359,13 @@ type Vault implements Node {
   Time-series of this vault's economic snapshots, oldest-first (ascending
   `timestamp`, matching `Account.statsHistory` — chart-ready, no consumer
   re-sort). Backs the vault dashboard's TVL / PnL charts.
+
+  `first` has no default on purpose: omitting it returns the whole bounded
+  daily series in one page, so the chart loads in a single request instead of
+  a chain of sequential paginated round-trips. (A literal `first` above
+  GRAPHQL_MAX_LIST_SIZE is still rejected pre-execution.)
   """
-  statsHistory(first: Int = 30, after: String, filter: VaultStatFilter): VaultStatConnection!
+  statsHistory(first: Int, after: String, filter: VaultStatFilter): VaultStatConnection!
 }
 
 """
--- a/packages/api/schema.v2.graphql
+++ b/packages/api/schema.v2.graphql
@@ -360,10 +360,11 @@ type Vault implements Node {
   `timestamp`, matching `Account.statsHistory` — chart-ready, no consumer
   re-sort). Backs the vault dashboard's TVL / PnL charts.
 
-  `first` has no default on purpose: omitting it returns the whole bounded
-  daily series in one page, so the chart loads in a single request instead of
-  a chain of sequential paginated round-trips. (A literal `first` above
-  GRAPHQL_MAX_LIST_SIZE is still rejected pre-execution.)
+  Omitting `first` returns up to a server-side cap (MAX_STATS_POINTS) in one
+  page, so the chart loads in a single request instead of a chain of paginated
+  round-trips — while still bounding the response for an append-only series.
+  Page forward with `after` if a vault ever exceeds the cap. (A literal `first`
+  above GRAPHQL_MAX_LIST_SIZE is still rejected pre-execution.)
   """
   statsHistory(first: Int, after: String, filter: VaultStatFilter): VaultStatConnection!
 }
--- a/packages/api/schema.v2.graphql
+++ b/packages/api/schema.v2.graphql
@@ -1601,6 +1601,12 @@ input ActivityFilter {
   conditionIds: [Bytes!]
   pickConfigId: Bytes32
 
+  """
+  Restrict activity to a single position token. With `account`, predictions
+  match only the side that minted this token; trades match exact token.
+  """
+  token: Address
+
   """Restrict to a subset of activity types. `[]` is a zero-result query."""
   types: [ActivityType!]
 
--- a/packages/api/schema.v2.graphql
+++ b/packages/api/schema.v2.graphql
@@ -1690,7 +1690,7 @@ type Protocol {
   Live, current protocol-wide stats — a single `ProtocolStat` with
   `timestamp = now`. Volume / trade-count / open-interest reflect the
   in-progress period and `totalValueLocked` is read across every configured
-  vault at query time. Use `statsHistory` for the recorded daily series.
+  vault at query time. Use `statsHistory` for the recorded snapshot series.
   """
   stats: ProtocolStat!
 
@@ -1699,8 +1699,18 @@ type Protocol {
   volume / TVL / open-interest time-series charts. (Previously named
   `stats`; renamed to `statsHistory` so `stats` can be the live singular,
   matching the `Vault` access pattern.)
+
+  Snapshots are recorded at a sub-daily cadence (configurable; ~4-hourly in
+  production), so the raw series runs to thousands of rows. Pass `interval`
+  to downsample server-side into fixed-width buckets — stock fields
+  (`cumulative*`, `openInterest`, `escrowBalance`, `totalValueLocked`) take
+  the bucket's last (latest) snapshot; flow fields (`periodVolume`,
+  `periodTradeCount`) sum across the bucket; the node's `timestamp` is the
+  bucket start. With no `interval` the raw per-snapshot series is returned.
+  The default page (`first` omitted) fits the whole bucketed series in one
+  request, so the analytics dashboard no longer paginates.
   """
-  statsHistory(first: Int = 100, after: String, filter: ProtocolStatFilter): ProtocolStatConnection!
+  statsHistory(interval: TimeInterval, first: Int = 100, after: String, filter: ProtocolStatFilter): ProtocolStatConnection!
   openInterestByCategory: [CategoryOpenInterest!]!
 
   """
--- a/packages/api/schema.v2.graphql
+++ b/packages/api/schema.v2.graphql
@@ -1719,8 +1719,9 @@ type Protocol {
   the bucket's last (latest) snapshot; flow fields (`periodVolume`,
   `periodTradeCount`) sum across the bucket; the node's `timestamp` is the
   bucket start. With no `interval` the raw per-snapshot series is returned.
-  The default page (`first` omitted) fits the whole bucketed series in one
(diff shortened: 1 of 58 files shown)


nostradamus-agent:
🔮 **Verdict: Request changes**

Two blockers before this staging → main promotion is merge-safe:

1. `packages/sdk/queries/vault.ts` adds `GET_VAULT_ACCOUNT_VALUE` with `statsHistory(interval: DAY, first: 366)`. `/v2/graphql` pre-execution validation rejects explicit `first > GRAPHQL_MAX_LIST_SIZE` (default 100), so the new vault AUM query will fail with `PAGINATION_LIMIT_EXCEEDED` before the resolver runs. Either omit `first` if you want the schema/resolver default, or page at `first <= 100` like `fetchVaultStats` already does.

2. `packages/app/src/components/vaults/pages/VaultsPageContent.tsx` now uses the indexed account value for AUM, but the deposit disabled state still treats missing `vaultAccountValue` as `tvlWei = 0n`. The UI hides the balance until `isBalanceReady`, but it can still enable/submit deposits before the AUM query resolves, undercounting capacity and letting an over-cap deposit through the client-side gate.

Also fix the generated SDK v2 types: `Protocol.statsHistory(interval:)` exists in the schema, but `ProtocolStatsHistoryArgs` in `packages/sdk/types/graphql.v2.ts` still lacks `interval`, so codegen is stale.

Validation I checked:
- GitHub checks/statuses are green now.
- Local merge-tree is clean; no textual conflicts.
- `git diff --check origin/main...HEAD` passes.
- Local Vitest could not run in the disposable worktree because `node_modules` is absent (`vitest: not found`); CI is the executable signal here.


_Note: inline review submission failed, so posting as a top-level review comment._

nostradamus-agent (changes requested):
🔮 **Verdict: Request changes**

Formal blocking review without inline anchors because GitHub could not resolve one inline line on the staging→main diff.

Blockers:
- `packages/sdk/queries/vault.ts:150` — `GET_VAULT_ACCOUNT_VALUE` uses `statsHistory(interval: DAY, first: 366)`, but `/v2/graphql` rejects explicit `first > GRAPHQL_MAX_LIST_SIZE` (default 100) before resolver execution. This will fail with `PAGINATION_LIMIT_EXCEEDED`. Omit `first` if relying on the default, or page at `first <= 100`.
- `packages/app/src/components/vaults/pages/VaultsPageContent.tsx:617` — capacity calculation falls back to `tvlWei = 0n` while `vaultAccountValue` is loading, but the deposit button is not disabled by `isBalanceReady`. This can undercount AUM and allow an over-cap submit through the client-side gate.

Non-blocking but should be fixed here: `packages/sdk/types/graphql.v2.ts` is stale; schema has `Protocol.statsHistory(interval:)`, but `ProtocolStatsHistoryArgs` lacks `interval`.

noahlitvin:
## Thorough review — Staging → Main (#1891)

Reviewed the full net diff (`git diff main...staging`): **20 files, +807/−106** across `api`, `app`, `sdk`. All three suites pass locally after building the SDK + generating Prisma: **SDK 725, API 856, app 765**.

### Scope note
The commit list shows `#1884` (market-keeper keyset relist pagination), but it **nets to zero** against `main` — that content was already merged via the prior "Staging → Main (#1885)" (it's present at the merge base). No keeper files appear in the net diff, so nothing keeper-side ships here. The real surface is four changes:

| PR | Change | Verdict |
|----|--------|---------|
| #1889 | Activity `token` filter (scope position activity by token) | ✅ Correct |
| #1886 | `Protocol.statsHistory` server-side downsampling + TVL/cross-family memoization | ✅ Correct |
| #1890 | Vault AUM from indexed account value | ✅ Correct |
| #1888 | Singles Vault tab | ✅ Correct |

---

### ✅ Verified correct

**Vault AUM ordering (#1890)** — `fetchVaultAccountValue` reads `statsHistory(interval: DAY, first: 366)` and takes `.at(-1)` as the latest bucket. I checked whether `first: 366` could return the *oldest* 366 days (making `.at(-1)` stale for accounts >1yr old). It can't: `getAccountStatsHistory` defaults the window to `[now − 365d, now]` (`timeSeriesQueries.ts:41-63`), a rolling grid that always ends at `now` and is sorted ascending (`accountStatsHistory.ts:139`). `first: 366` (= `MAX_STATS_POINTS`) covers the whole window, so `.at(-1)` is always today's bucket regardless of account age. The three AUM terms (wallet balance + deployed + claimable) now share one freshness boundary — a genuine improvement over the old live-contract + cached-snapshot mix.

**Activity token scoping (#1889)** — The refactor from single `where` objects to `{ AND: clauses }` is sound. The account+token prediction case correctly matches only the minted side (`predictor AND predictorToken` OR `counterparty AND counterpartyToken`), avoiding the opposite-side leak. Trade-side token∩conditionTokens intersection uses `{ token: { in: [] } }` as a verified match-nothing sentinel. Field names (`Picks.predictorToken`/`counterpartyToken`, `SecondaryTrade.token`, `Prediction.pickConfiguration`) all check out against `schema.prisma`.

**Protocol downsampling + memoization (#1886)** — `singleFlightByChain` caches the in-flight promise keyed by chainId, evicts on rejection — same pattern as the v1 `protocolStatsCache`. Keying on chainId alone is correct since `computeProtocolTvl`/`crossFamilyAvailableByRawTs` take only chainId (independent of statsHistory filter/pagination). Bucketing is deterministic (stock fields → bucket's last row; flow fields summed; label = bucket start), so the index-based cursor stays stable within the 60s window. SDK sends `first: null` to get the full bucketed series in one page, with `GET_PROTOCOL_STATS_HISTORY_PAGE` retained as a pagination safety net.

---

### Minor (non-blocking)

1. **Redundant AND clause** — `activity.ts`: when both `args.filter.pickConfigId` and `conditionPickConfigIds` are set, two `{ pickConfigId: ... }` clauses get pushed into the same `AND`. Since the conditionPickConfigIds branch already filters `ids` to match `pickConfigId`, the two are identical — harmless but redundant. The old code overwrote; behavior is equivalent.

2. **Test gaps** — `activity.test.ts` covers the account+token side-matching case well, but doesn't exercise: token-only (no account), and the token∩conditionTokens intersection (both the in-set `{ token }` and not-in-set `{ in: [] }` branches). The `{ in: [] }` sentinel in particular is logic worth a direct test. Not blocking given the existing coverage of the core path.

---

### Verdict
**Ready to merge.** No correctness issues found in the net diff; the one ordering risk I chased down is provably safe given the rolling-window default. The two minor items are cleanup, not blockers.

noahlitvin:
## Review blockers addressed — commit `b4aac3f33`

Three issues raised in follow-up review were confirmed legitimate and fixed on `staging`:

1. **`first: 366` would 400 in production (blocker).** The vault-AUM query sent a literal `first: 366`; `queryValidation.ts:43-66` rejects any literal `first`/`take`/`limit` above `GRAPHQL_MAX_LIST_SIZE` (default **100**) with `PAGINATION_LIMIT_EXCEEDED`. The SDK unit test mocked the response, so it never exercised validation. **Fixed** by dropping the explicit `first` — `Account.statsHistory` defaults to its `MAX_STATS_POINTS` (366) cap, returning the full rolling `[now−365d, now]` daily grid, so `.at(-1)` is still today's bucket. (Same approach the Protocol query uses via a null variable.) Test now asserts no literal `first:`.

2. **Deposit cap race (footgun).** The deposit button wasn't gated on `isBalanceReady`; while indexed AUM loads, `tvlWei` reads `0`, so `exceedsVaultCapacity` understates the total and a near-cap vault could let an over-cap deposit through the client check. **Fixed** by gating the capacity clause on `isBalanceReady`.

3. **Stale generated types.** `ProtocolStatsHistoryArgs` lacked the `interval` arg the SDL exposes. **Fixed** by rerunning `generate-types` (only adds `interval?: InputMaybe<TimeInterval>` + doc comments).

Also folded the redundant duplicate `pickConfigId` AND-clause in the activity resolver into a single branch (no behavior change).

**Verification:** SDK 725 / API 856 / app 765 tests pass; `api` and `app` `type-check` both clean; lint-staged (eslint + prettier) clean.

nostradamus-agent (changes requested):
🔮 **Verdict: Request changes — not safe to merge yet.**

Two blockers before this staging → main promotion is merge-safe:

1. **GitHub reports the PR as dirty / not mergeable.** `origin/main` has one main-only commit (`4f020958d Staging -> Main (#1887)`), and the direct merge conflicts in `packages/sdk/queries/protocol.ts`. This cannot merge as-is.

2. **`Protocol.statsHistory` public GraphQL contract silently truncates normal callers.** The SDL documents that omitting `first` returns the whole bucketed series, but the field still declares `first: Int = 100`. Apollo applies that default before the resolver, so `args.first` is `100`, not nullish, and `Protocol.ts` only uses `nodes.length` when `args.first ?? nodes.length` is nullish. The SDK works around this with `first: null`, but direct GraphQL clients omitting `first` will only get 100 rows despite the docs. Fix by removing the SDL default or making the contract/docs explicitly require `first: null` for full bucketed history.

Reviewed areas that look okay after the previous fixes: vault tab data isolation, Singles Vault tab exposure, indexed AUM capacity guard, PnL chart loading, position activity token scoping, vault stats pagination, and market-keeper keyset pagination / alert reporting.

Validation notes: GitHub/Railway/Vercel statuses on the staging head are green, but local targeted test runs hit runtime `ENOSPC` inode exhaustion after/while attempting package checks, so I’m not treating local tests as a clean gate. `git diff --check origin/main...origin/staging` passed in the keeper review path.

PR #1895, opened 2026-06-19
perf(vaults): load PnL chart series in one request instead of N paginated round-trips

noahlitvin:
## Problem

The chart on `/vaults` is slow to load. Each vault's stats series is fetched by paginating `vault.statsHistory` at `pageSize=100`. Production vaults carry **~1251 daily snapshots**, so `fetchVaultStats` issued **~13 sequential GraphQL round-trips per vault** (~4.5–5.0s of serial fetching each), and the resolver re-ran an **unbounded** `findMany` (fetch-all → dedupe → slice) on *every* page.

The forcing function was the schema default `statsHistory(first: Int = 30)`: omitting `first` still returned only 30 rows, so the SDK had no choice but to page.

## Fix

The series is a bounded daily set, so return it whole in a single request — mirroring how `Account.statsHistory` / `fetchVaultAccountValue` already work:

- **schema** (`schema.v2.graphql`): drop the `first: Int = 30` default on `Vault.statsHistory` → `first: Int`. Omitting `first` now returns the full deduped series in one page; an explicit `first` still pages and is capped by `GRAPHQL_MAX_LIST_SIZE`.
- **resolver** (`Vault.ts`): when `first` is null, page size = `totalCount` (whole series).
- **sdk** (`vault.ts`): `fetchVaultStats` issues one no-`first` request — pagination loop removed.

## Verification

Tested against the live API (production read replica):

| | before | after |
|---|---|---|
| Requests per vault | ~13 (sequential) | **1** |
| Wall-clock per vault | ~4.5–5.0s | **~0.66s** |
| Points returned | 1251 (accumulated) | 1251 (one page) |

Tests updated test-first (red → green); `Vault.test.ts` and `sdk vault.test.ts` cover the new single-request behavior. SDK + API + App type-checks pass.

### DIFF
--- a/packages/api/src/graphql/v2/resolvers/Vault.ts
+++ b/packages/api/src/graphql/v2/resolvers/Vault.ts
@@ -181,10 +181,6 @@ export const Vault: VaultResolvers = {
 
   statsHistory: async (parent, args) => {
     const row = parent as unknown as VaultRow;
-    const first = clampTake(args.first ?? 30, {
-      defaultTake: 30,
-      maxTake: 365,
-    });
     const after = args.after ? decodeCursor(args.after) : null;
     const skip = after && /^\d+$/.test(after.k) ? Number(after.k) + 1 : 0;
 
@@ -249,6 +245,14 @@ export const Vault: VaultResolvers = {
       (a, b) => a.timestamp - b.timestamp
     );
     const totalCount = deduped.length;
+    // No `first` => return the whole bounded daily series in one page (mirrors
+    // Account.statsHistory), so the SDK loads the chart in a single request
+    // instead of a chain of sequential paginated round-trips. An explicit
+    // `first` still pages and is capped (also by the global list-size limit).
+    const first =
+      args.first == null
+        ? totalCount
+        : clampTake(args.first, { defaultTake: 30, maxTake: 365 });
     const pageRows = deduped.slice(skip, skip + first + 1);
 
     return buildConnection({
--- a/packages/api/src/graphql/v2/resolvers/Vault.ts
+++ b/packages/api/src/graphql/v2/resolvers/Vault.ts
@@ -35,6 +35,14 @@ export type VaultRow = {
   chainId: number;
 };
 
+// Upper bound on rows returned by `statsHistory` in a single page when the
+// caller omits `first`. The vault snapshot table is append-only (grows ~1/day),
+// so an unbounded "return everything" page would balloon the public response
+// over time. This cap keeps the common case a single request (current series
+// is well under it) while bounding the worst case; a vault that ever exceeds
+// the cap pages forward via the offset `after` cursor (see SDK fetchVaultStats).
+export const MAX_STATS_POINTS = 2000;
+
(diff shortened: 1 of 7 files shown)


packages/api/src/graphql/v2/resolvers/Vault.ts
```diff
@@ -249,6 +245,14 @@ export const Vault: VaultResolvers = {
       (a, b) => a.timestamp - b.timestamp
     );
     const totalCount = deduped.length;
+    // No `first` => return the whole bounded daily series in one page (mirrors
+    // Account.statsHistory), so the SDK loads the chart in a single request
+    // instead of a chain of sequential paginated round-trips. An explicit
+    // `first` still pages and is capped (also by the global list-size limit).
+    const first =
```
nostradamus-agent: This is the problematic bit: omitted `first` becomes `totalCount`, so the public API can return every historical vault snapshot in one response and bypass the explicit `first` cap / `GRAPHQL_MAX_LIST_SIZE` validation. The resolver already fetched all DB rows before this, and the SDK now always uses this uncapped path with no `pageInfo` fallback. Please put a real server-side cap/window on the omitted-`first` path, or keep SDK pagination while optimizing the DB side.

nostradamus-agent (changes requested):
Reviewed PR #1895. Requesting changes.

Blocking issue:
- `Vault.statsHistory` now treats omitted `first` as `totalCount`, so `statsHistory { nodes { ... } }` bypasses both the old 30-row default and the explicit `first` cap / `GRAPHQL_MAX_LIST_SIZE` validation. The resolver still fetches every matching `protocolStatsSnapshot` row and dedupes in memory before slicing. The SDK now always sends the unbounded form and removed the pageInfo fallback. That makes the fast path work for today's ~1.2k rows, but it also turns a public GraphQL field into an uncapped response with underestimated query complexity. Please add a real server-side cap/window for the omitted-`first` path, or keep SDK pagination as a fallback while optimizing the resolver query.

Non-blocking:
- `fetchVaultStats(address, chainId, pageSize?)` effectively removed the optional third arg. Runtime JS is fine, but TS SDK consumers passing `pageSize` will get an arity/type error; preserve it or document the breaking change if this is public surface.

Checks:
- Inspected diff against the PR base `staging` SHA.
- `git diff --check` passed.
- GitHub status is green for Vercel app/docs.
- Local package tests were not runnable here because this clone has no `node_modules`; I did not install dependencies in the dirty shared repo.

nostradamus-agent (approved):
Approved on re-review.

The previous blocker is fixed: omitted `first` is now capped server-side (`MAX_STATS_POINTS`) instead of using `totalCount`, and the SDK restored `pageInfo`/`after` pagination so history remains complete if the cap is ever exceeded. The optional `pageSize` argument is also preserved.

Checks:
- Reviewed delta from `beeb9cef` to `2351ea2`
- Re-checked full PR diff against `staging`
- `git diff --check` passed
- GitHub status is green for Vercel app/docs

Note: local package tests were not runnable in this shared clone because `node_modules` is absent; I did not install dependencies into the dirty repo.

PR #1897, opened 2026-06-22
fix(vaults): correct inflated vault balance (deployed over-counting)

noahlitvin:
## Problem

The `/vaults` headline **VAULT BALANCE** has been inflated since #1890 ("use indexed vault account value for vaults AUM"):

- **Core**: showed **40,942.68** (was ~$35k before the merge)
- **Edge**: showed **4,659.07** — but only ~$2k was ever deposited and the PnL chart shows ~$900 profit (true ≈ $2.9k)

## Root cause

`fetchVaultAccountValue` summed `collateralBalance + deployedCollateral + claimableCollateral` from the **`account(...)`** surface. The two surfaces compute `deployedCollateral` differently:

- `Account.statsHistory.deployedCollateral` (`timeSeriesQueries.queryAccountBalance`) keys "still open" off **`Prediction.settledAt`**. Losing predictions are frequently never settled on-chain, so their counterparty collateral stays counted as deployed **forever**.
- `Vault.stats.deployedCollateral` (`fetchVaultDeployed`) keys it off **`Picks.resolved/resolvedAt`** — resolved positions drop out immediately. This is the figure the PnL chart already plots.

So the account path piled phantom "deployed" on top of the (correct) liquid balance.

## Verification (chain 5064014)

On-chain reads confirm the **liquid** term was already correct — only **deployed** was wrong:

| | balanceOf == availableAssets == account.collateralBalance | account deployed (wrong) | vault deployed (correct) | cumulativePnl |
|---|---|---|---|---|
| Core | 33,141.59 | 7,310.49 | 2,105.83 | 1,869.42 |
| Edge | 2,339.57 | 2,308.18 | 522.73 | 952.77 |

Resulting balance:
- **Core: 40,942.68 → ~35,362** (≈ pre-merge $35k) ✓
- **Edge: 4,659.07 → ~2,957** (= $2k deposit + ~$952 PnL) ✓

## Fix

Repoint `GET_VAULT_ACCOUNT_VALUE` / `fetchVaultAccountValue` to the vault entity's live `stats` snapshot (`balance + deployedCollateral + claimableCollateral`). The hook's return shape is unchanged, so the `VaultsPageContent` component (which already renders deployed as a subset of total) needs no change.

## Tests

- Updated `packages/sdk/queries/__tests__/vault.test.ts` to assert the query reads `vault { stats }` (regression guard against the `account(...)` source) and to cover null-vault / null-stats fallbacks.
- SDK: 728 passed. App: 766 passed (`VaultsPageContent` suite included).

### DIFF
--- a/packages/sdk/queries/__tests__/vault.test.ts
+++ b/packages/sdk/queries/__tests__/vault.test.ts
@@ -189,40 +189,32 @@ describe('fetchVaultStats', () => {
 });
 
 describe('GET_VAULT_ACCOUNT_VALUE document', () => {
-  test('queries account collateralBalance and account statsHistory from one v2 surface', () => {
+  test('reads the live vault `stats` snapshot, not the account surface', () => {
+    // Must source from the vault entity: its `deployedCollateral` is keyed off
+    // Picks.resolved/resolvedAt, so resolved-but-unsettled (losing) positions
+    // drop out. The `account(...)` surface keys deployed off Prediction.settledAt
+    // — losing predictions are never settled on-chain, so they stay counted as
+    // deployed forever and inflate the balance. Regression guard for that.
     expect(GET_VAULT_ACCOUNT_VALUE).toContain(
-      'account(address: $address, chainId: $chainId)'
+      'vault(address: $address, chainId: $chainId)'
     );
-    expect(GET_VAULT_ACCOUNT_VALUE).toContain('collateralBalance');
-    expect(GET_VAULT_ACCOUNT_VALUE).toContain('amount');
-    expect(GET_VAULT_ACCOUNT_VALUE).toContain('statsHistory(interval: DAY)');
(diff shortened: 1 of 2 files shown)


nostradamus-agent (approved):
🔮 Reviewed. Ship it.

What I checked:
- `fetchVaultAccountValue` now uses `vault(address).stats` instead of `account(...).statsHistory`, so the displayed balance is sourced from the same vault snapshot family as the PnL/TVL chart.
- The query handles `vault: null`, `stats: null`, and BigInt scalars serialized as strings/numbers.
- Live v2 GraphQL accepted the new `vault { stats { timestamp balance deployedCollateral claimableCollateral } }` selection for the Core Vault.
- Focused SDK test file ran green: `queries/__tests__/vault.test.ts` showed 13/13 passing. Caveat: the local runner then failed while writing Vitest cache because this sandbox hit inode exhaustion (`node_modules/.vite`), not because the tests failed.

No blocking findings.

PR #1901, opened 2026-06-24
fix(api): drop resolved-but-unsettled losses from deployedCollateral

noahlitvin:
## Problem

`Account.statsHistory.deployedCollateral` — and the v1 `accountBalance` query, since both share the `queryAccountBalance` SQL in `packages/api/src/services/timeSeriesQueries.ts` — decided whether collateral was "still deployed" off `Prediction.settledAt`.

`Prediction.settledAt` and `PickConfiguration.resolved` are set by **decoupled** flows: `settledAt` only on the on-chain `PredictionSettled` event, `Picks.resolved`/`result` as soon as the conditions settle (early resolution). A losing predictor has no incentive to ever settle on-chain, so `COUNTERPARTY_WINS` losses sat with `settledAt = null` and were counted as deployed **forever**, inflating the headline balance well above true AUM.

Example: kernel `0xF20e8BA6F2b84D0d48865A47E8832494f2f6F34F` showed ~$25 deployed when only ~$5 was genuinely open — the ~$20 gap was resolved-but-unsettled losses.

The codebase already documented this divergence in `packages/sdk/queries/vault.ts`, and the vault path (`vaultAggregator.deployedAt`) already does it correctly.

## Fix

Key the Prediction branch's deploy window off `Picks.resolved`/`resolvedAt` instead of `settledAt`, mirroring `vaultAggregator.deployedAt`:

- `pickConfigId` null → stays deployed (no resolution signal; rare RPC-error case)
- pickConfig unresolved → deployed
- pickConfig resolved → drops at `resolvedAt`, regardless of `settledAt`

The legacy `position` branch keeps its `settledAt` window (no pickConfig; settles as a unit). One shared SQL change fixes both `/graphql` (v1) and `/v2/graphql` (v2).

## Test

Adds a real-SQL integration scenario to `timeSeriesQueries.balance.integration.test.ts`: a resolved-but-unsettled loss (20) + an open position (5). Asserts deployed reads 25 while both are live, then drops to 5 once the pickConfig's `resolvedAt` passes. Verified red before the fix, green after. Related suites (timeSeriesQueries unit/pnl, accountStatsHistory, Vault) all pass.

## Note

A `PREDICTOR_WINS` position now leaves "deployed" at `resolvedAt` but only enters "claimable" at `settledAt` (claimable amounts aren't known until the settle event), so it's briefly in neither bucket — same as the vault path, and short-lived in practice. Attribution-on-transfer is a separate pre-existing limitation, tracked separately.

### DIFF
--- a/packages/api/src/services/timeSeriesQueries.balance.integration.test.ts
+++ b/packages/api/src/services/timeSeriesQueries.balance.integration.test.ts
@@ -36,13 +36,18 @@ const HOLDER_OTHER_SIDE = '0xbbbb000000000000000000000000000000000002';
 // Predictor who redeems only PART of a winning side — claimable must decrement
 // by the redeemed amount, not drop to zero.
 const HOLDER_PARTIAL_CLAIM = '0xcccc000000000000000000000000000000000003';
+// Predictor holding one LOST position (pickConfig resolved COUNTERPARTY_WINS but
+// the prediction was never settled on-chain, settledAt = null) plus one still-
+// open position. Deployed collateral must drop the lost leg at the pickConfig's
+// resolvedAt, not wait forever on settledAt.
+const HOLDER_DEPLOY = '0xdddd000000000000000000000000000000000004';
 
 // Minimal slices of the tables queryAccountBalance touches. `position` (the V1
 // legacy table) exists only so the all_deployed UNION branch resolves; it stays
 // empty. Prediction carries both predictionId and pickConfigId so the DDL is
 // valid against the pre-fix query too (proving the test fails before the fix).
 const DDL = `
-  CREATE TABLE "Picks" (id text PRIMARY KEY, "predictorToken" text, "counterpartyToken" text);
+  CREATE TABLE "Picks" (id text PRIMARY KEY, "predictorToken" text, "counterpartyToken" text, resolved boolean, "resolvedAt" integer);
   CREATE TABLE "Prediction" (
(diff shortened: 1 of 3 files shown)


nostradamus-agent (approved):
🔮 Reviewed. Ship it.

What I checked:
- `queryAccountBalance` now ends the prediction deploy window on `Picks.resolvedAt` instead of `Prediction.settledAt`, while preserving the legacy `position` branch on `settledAt`.
- Null/missing pick-config behavior matches the documented vault aggregator semantics: `pickConfigId = null` stays deployed; missing/resolved-without-resolvedAt drops immediately.
- Regression coverage exercises the actual SQL shape for resolved-but-unsettled losses.

Verification:
- `git diff --check origin/staging...HEAD`
- `pnpm install --frozen-lockfile`
- `pnpm --filter @sapience/api run prisma:generate`
- `pnpm --filter @sapience/api exec vitest run --maxWorkers=1 --minWorkers=1 src/services/timeSeriesQueries.balance.integration.test.ts src/services/timeSeriesQueries.test.ts src/graphql/v2/resolvers/queries/accountStatsHistory.test.ts src/graphql/v2/resolvers/Vault.test.ts` → 38 passed, 4 skipped (DB-gated integration skipped locally)
- targeted ESLint + Prettier on changed API/SDK files

CI visible from GitHub is green for Vercel app/docs. No blockers found.

PR #1910, opened 2026-06-27
fix(market-keeper): migrate Discovery market fetch to /markets/keyset

noahlitvin:
## Problem

Prod and staging **Discovery** keepers crash on `market-keeper:generate`:

```
[Polymarket] Page 235: Fetched 100 markets (max endDate: 2026-07-06T10:00:00.000Z)
[Polymarket API] Failed to fetch ending-soon markets: HTTP 422 Unprocessable Entity
[Polymarket API] Response body: {"type":"validation error","error":"offset too large, use /markets/keyset for deeper pagination"}
```

Polymarket deprecated the legacy Gamma `/markets` and `/events` endpoints (sunset 2026-05-01) and now hard-rejects deep offset pagination.

## Root cause

`fetchEndingSoonestMarkets` (`packages/market-keeper/src/generate/market.ts`) paginated `/markets` primarily by advancing `end_date_min`, but fell back to `limit=100&offset=...` whenever every market in a batch shared a single `endDate` (cursor couldn't advance). A large cluster of markets all closing at one round timestamp (e.g. `2026-07-06T10:00:00Z`, visible as pages 233–235 pinned to the same `endDate` in the logs) drove `offset` past Gamma's newly-enforced cap → HTTP 422 → process exits 1 → `start:discovery` fails.

This is unique to the Discovery generate step, which is why Settlement / Market Data / Metadata are unaffected.

## Fix

Replace offset pagination with cursor-based **keyset pagination** (`after_cursor` / `next_cursor`) on `/markets/keyset`, bounding **both** ends of the window server-side via `end_date_min` / `end_date_max`. Keyset has no offset cap, so same-`endDate` clusters of any size are walked without truncating — fixing both the crash and the latent silent-truncation risk that a naive "cap the offset" fix would have introduced (it would silently drop markets in any cluster larger than the cap).

- Extracted the pagination into `fetchEndingSoonMarketsViaKeyset(minEndDate, maxEndDate)`.
- Preserves **ending-soonest** order via `order=endDate&ascending=true` (Gamma only honors `ascending` when `order` is set, so both are required for a front-to-back window walk).
- **Stuck-cursor guard:** stops when the server returns an unchanged `next_cursor` (the [reported keyset cursor-ignored bug](https://github.com/Polymarket/agents/issues/227)). This is detected directly rather than inferred from a zero-new-market page — a page can legitimately add zero markets because every row was deduped or dropped by the window guard, and stopping on that would truncate the walk early. A `MAX_PAGES` backstop guards against any other non-termination.

### Supplementary `/events` → `/events/keyset`

The supplementary `fetchMarketsByEventTags` (used to backfill markets the `/markets` list misses) called the legacy `/events` endpoint (`limit=200`, unpaginated), which is on the same deprecation track. Migrated it to `/events/keyset` with cursor pagination: response shape is `{ events, next_cursor }`, bounded by `end_date_max`, following `next_cursor` per tag with the same stuck-cursor + `MAX_PAGES` guards. Behavior is otherwise unchanged — it still doesn't dedup (the caller dedups via `seenConditionIds`).

## API contract verification

The keyset contract was confirmed against [the official keyset docs](https://docs.polymarket.com/api-reference/markets/list-markets-keyset-pagination) **and** live read-only probes of the real API:

- `/markets/keyset` accepts `after_cursor`, `order=endDate`, `ascending=true`, `end_date_min/max`; returns `{ markets, next_cursor }` with `next_cursor` omitted on the final page; rejects `offset` with 422; cursor advances and results come back soonest-first.
- `/events/keyset` returns `{ events, next_cursor }`, events carry `markets[]`, and `tag_slug` filters correctly.

## Tests

New `src/__tests__/fetch-ending-soonest-keyset.test.ts` (16 tests, TDD red→green):

**Markets walk**
- hits `/markets/keyset`, never sends `offset=`
- bounds the window with `end_date_min` / `end_date_max`
- preserves `order=endDate&ascending=true`
- follows `next_cursor` across pages as `after_cursor`
- **regression:** paginates a large same-`endDate` cluster purely by cursor (the 422 scenario)
- dedups by `conditionId`; boundary guard at `maxEndDate`; error surfacing
- stops on an unchanged cursor (cursor-stuck guard)
- does **not** stop on a zero-new page while the cursor is still advancing

**Events walk**
- hits `/events/keyset`, never sends `offset`, bounds with `end_date_max`
- flattens `event.markets` and injects the parent event for grouping
- follows `next_cursor` across event pages per tag
- drops markets at/beyond `maxEndDate`
- skips a tag on a non-ok response and continues to the next tag
- stops on an unchanged cursor

Full `@sapience/market-keeper` suite passes (546 passed / 4 skipped); `tsc` build and eslint clean.

### DIFF
--- a/packages/market-keeper/src/generate/market.ts
+++ b/packages/market-keeper/src/generate/market.ts
@@ -8,34 +8,44 @@ import { fetchWithRetry } from '../utils';
 import { runPipeline, printPipelineStats, MARKET_FILTERS } from './pipeline';
 
 /**
- * Fetch markets that end soonest (for --ending-soon mode)
- * Orders by endDate ascending, no volume sorting
- * Iteratively fetches pages by moving end_date_min based on max endDate from previous response
+ * Page through Gamma's /markets/keyset endpoint for markets ending within the
+ * [minEndDate, maxEndDate) window, deduplicated by conditionId.
+ *
+ * Uses cursor-based pagination (after_cursor / next_cursor). Polymarket
+ * deprecated offset pagination on the legacy /markets endpoint (sunset
+ * 2026-05-01) and now returns HTTP 422 — "offset too large, use /markets/keyset
+ * for deeper pagination" — once the offset grows large. Discovery hit this
+ * whenever a large cluster of markets shared a single endDate: the old fetcher's
+ * offset fallback walked deep enough to trip the cap and the keeper crashed.
+ * Keyset has no offset cap, so it walks same-endDate clusters of any size
+ * without truncating.
+ *
+ * Both ends of the window are bounded server-side via end_date_min/end_date_max
+ * so the cursor walk terminates at the window edge instead of paging the entire
+ * active-market set.
  */
-export async function fetchEndingSoonestMarkets(): Promise<PolymarketMarket[]> {
-  // Minimum end time: current time + 1 minute (ISO format for API)
-  let currentMinEndDate = new Date(Date.now() + 60 * 1000).toISOString();
-  // Maximum end time: current time + MAX_END_DATE_DAYS
-  const maxEndDate = new Date(
-    Date.now() + MAX_END_DATE_DAYS * 24 * 60 * 60 * 1000
-  );
-
+export async function fetchEndingSoonMarketsViaKeyset(
+  minEndDate: Date,
+  maxEndDate: Date
+): Promise<PolymarketMarket[]> {
   const allMarkets: PolymarketMarket[] = [];
-  const seenConditionIds = new Set<string>(); // Track seen markets to deduplicate
-  // Polymarket's /markets endpoint silently caps `limit` at 100 server-side;
-  // requesting more returns 100 anyway, and the `markets.length < PAGE_SIZE`
-  // short-page stop signal below would then misfire after page 1.
-  const PAGE_SIZE = 100;
-  let pageCount = 0;
-  let offset = 0;
+  const seenConditionIds = new Set<string>(); // deduplicate across pages
+  const PAGE_SIZE = 100; // keyset max page size (since 2026-05-14)
 
-  console.log(
-    `[Polymarket] Fetching ending-soon markets until ${maxEndDate.toISOString()}...`
-  );
+  const baseParams =
+    `limit=${PAGE_SIZE}` +
+    `&active=true&closed=false` +
+    `&end_date_min=${encodeURIComponent(minEndDate.toISOString())}` +
+    `&end_date_max=${encodeURIComponent(maxEndDate.toISOString())}`;
+
+  let cursor: string | undefined;
+  let pageCount = 0;
 
   while (true) {
     pageCount++;
-    const url = `https://gamma-api.polymarket.com/markets?limit=${PAGE_SIZE}&offset=${offset}&active=true&closed=false&order=endDate&ascending=true&end_date_min=${currentMinEndDate}`;
+    const url =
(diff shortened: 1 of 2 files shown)


packages/market-keeper/src/generate/market.ts
```diff
@@ -8,34 +8,44 @@ import { fetchWithRetry } from '../utils';
...
-  const PAGE_SIZE = 100;
-  let pageCount = 0;
-  let offset = 0;
+  const seenConditionIds = new Set<string>(); // deduplicate across pages
+  const PAGE_SIZE = 100; // keyset max page size (since 2026-05-14)
 
-  console.log(
-    `[Polymarket] Fetching ending-soon markets until ${maxEndDate.toISOString()}...`
-  );
+  const baseParams =
```
nostradamus-agent, line 43: ⚠️ This needs to preserve the old `order=endDate&ascending=true` params. The keyset docs expose `order` separately and say `ascending` is only used when `order` is set; without it this no longer guarantees an ending-soonest walk/return order. It also makes the zero-new-page stop below unsafe if an unordered page contains only boundary/duplicate rows while later cursor pages still contain valid markets.

packages/market-keeper/src/generate/market.ts
```diff
@@ -53,66 +63,72 @@ export async function fetchEndingSoonestMarkets(): Promise<PolymarketMarket[]> {
...
-    if (
-      markets.length < PAGE_SIZE ||
-      maxEndDateInBatch >= maxEndDate ||
-      newMarketsCount === 0
-    ) {
+    // Safety net for the reported keyset bug where the server ignores the
+    // cursor and re-serves the first page with a persistent next_cursor: a page
+    // that adds zero new markets means we are not advancing — stop rather than
+    // loop forever.
+    if (newMarketsCount === 0) {
```
nostradamus-agent: ⚠️ `newMarketsCount === 0` is too broad as a stuck-cursor guard unless the endpoint is definitely ordered by `endDate`. A page can add zero because every row was filtered/deduped, not because the cursor is stuck; then we truncate the rest of the cursor stream. Safer pattern is the one in `relist/market.ts`: track repeated `next_cursor` values (plus a max-pages backstop), and keep walking when the cursor advances.

nostradamus-agent (changes requested):
Reviewed the market-keeper keyset migration. Requesting changes for one pagination correctness issue.

The focused test, typecheck, and lint are green locally, but the new request drops the old `order=endDate` ordering while relying on page order/zero-new-page behavior for safe pagination. Docs say `ascending` only applies when `order` is set, so this should keep the old `order=endDate&ascending=true` contract before merge.

nostradamus-agent (approved):
Re-reviewed latest commit `291ec22d6`. The previous pagination blocker is fixed: the markets keyset walk now restores `order=endDate&ascending=true`, the stuck-cursor guard checks repeated cursors instead of `newMarketsCount === 0`, and the zero-new advancing-cursor case is covered.

Also reviewed the new `/events/keyset` supplementary tag migration against Polymarket docs (`tag_slug`, `end_date_max`, `events`/`next_cursor` envelope) and the caller-side dedup keeps duplicate event-page markets harmless.

Checks run:
- `pnpm --filter @sapience/market-keeper exec vitest run src/__tests__/fetch-ending-soonest-keyset.test.ts --maxWorkers=1 --minWorkers=1` (16 passed)
- `pnpm --filter @sapience/market-keeper run type-check`
- `pnpm --filter @sapience/market-keeper run lint`

PR #1917, opened 2026-07-01
feat(api): reduce GraphQL page-size cap 100→25

7ayd:


packages/app/src/hooks/graphql/usePickConfigsByTokens.ts
```diff
@@ -20,7 +20,7 @@ export const PICK_CONFIGS_BY_TOKENS_QUERY = `
   query PickConfigsByTokens($tokens: [Address!]) {
     pickConfigurations(
       filter: { tokens: $tokens }
-      first: 100
+      first: 25
```
nostradamus-agent: This avoids the validation error, but it turns the hook into a silent truncation point. `useEnrichedTokens()` can pass tokens from all fetched secondary trades/listings; if those map to more than 25 pick configs, everything after the first page is missing enrichment. Please add cursor pagination (`pageInfo/endCursor`) or chunk the token lookup so all requested tokens are covered.

packages/api/src/core/config.ts
```diff
@@ -36,7 +36,7 @@ const validators = {
     desc: 'TTL (ms) for entries in the v2 Automatic Persisted Queries cache. After this window an unrecognized hash forces clients to resend the full query. Setting too low defeats the purpose of APQ; setting too high lets unused queries linger in memory.',
   }),
   GRAPHQL_MAX_LIST_SIZE: num({
-    default: 100,
+    default: 25,
```
nostradamus-agent, line 39: Dropping the global cap to 25 exposes remaining callers that still send larger variable values. I found at least `useAccuracyLeaderboard(100)` and market-keeper cleanup’s `conditions(first: chunk.length)` where `chunk.length` is 50; both will now fail validation before reaching resolvers.

nostradamus-agent (changes requested):
Verdict: **request changes** — the cap drop itself is fine, but a few callers still either exceed 25 or now silently truncate.

Blocking:
- `packages/sdk/queries/leaderboard.ts:59-168` / `packages/app/src/components/leaderboard/Leaderboard.tsx:320-321` — accuracy leaderboard still sends `first: 100` via `useAccuracyLeaderboard(100)`. With `GRAPHQL_MAX_LIST_SIZE=25`, query validation rejects it before resolver execution. Page it like the PNL leaderboard, or intentionally reduce the UI to 25.
- `packages/app/src/hooks/graphql/usePickConfigsByTokens.ts:21-56` — changing `pickConfigurations(first: 100)` to `first: 25` makes token enrichment silently incomplete once the trade/listing token set maps to >25 pick configs. This is especially visible now that secondary trades fetch all pages. Add `after`/`pageInfo` and loop, or chunk so every requested token can be enriched.
- `packages/market-keeper/src/cleanup/api.ts:90,181-193` — cleanup still chunks ids by 50 and calls `conditions(first: chunk.length, filter: { conditionIds: chunk })`. Any full chunk now sends `first: 50` and gets `PAGINATION_LIMIT_EXCEEDED`; setting only `first: 25` would also miss ids 26-50, so split that GraphQL lookup into <=25-id chunks.

Checked: `git diff --check origin/staging...review/pr-1917` is clean. I did static review plus delegated focused passes; local focused Vitest was not runnable in the available worktree because test deps/binaries are missing (`vitest`/`tsup` not found).

nostradamus-agent (approved):
Verdict: **approved** — the three blockers from my prior review are fixed.

Verified:
- Accuracy leaderboard now pages with `first: 25` + `after` and still returns up to the requested display limit.
- `usePickConfigsByTokens` now requests `first/after`, reads `pageInfo`, and loops through the connection, so enrichment no longer silently truncates at 25.
- Market-keeper cleanup now splits GraphQL id-filter chunks at `GRAPHQL_PAGE_SIZE` while keeping REST admin batches at 50.
- No explicit `first`/`take`/`limit` literals above 25 remain in non-test files touched by this PR.
- `git diff --check origin/staging...review/pr-1917` is clean.
- Ran SDK focused tests: `pnpm --filter @sapience/sdk exec vitest run queries/__tests__/leaderboard.test.ts queries/__tests__/pagination.test.ts --pool=threads --poolOptions.threads.singleThread` — 36 passed.

Caveat: I tried the app/keeper focused tests too, but the disposable worktree had symlinked deps because `/opt/data` is inode-starved; Vite resolved `@sapience/sdk/*` through the wrong workspace path and failed before test execution. I’m treating that as local sandbox wiring, not a PR failure.

PR #1918, opened 2026-07-01
fix docker

7ayd:


nostradamus-agent (approved):
Approved. This directly fixes the Railway Docker build failure: the Dockerfile no longer COPYs packages/edge-cache/package.json, which is absent from the PR/staging tree. I also verified every remaining Dockerfile COPY source exists at this head SHA, and current GitHub checks are green.

PR #1920, opened 2026-07-01
update docker

7ayd:


nostradamus-agent (approved):
Approved. Same minimal Dockerfile fix as the staging PR: removes the stale COPY packages/edge-cache/package.json line, and packages/edge-cache/package.json is absent from the main PR tree. I verified all remaining Dockerfile COPY sources exist at this head SHA and there are no remaining edge-cache references in the Dockerfile. Vercel checks were still pending at review time, but this Dockerfile-only change is correct for the Railway cache-key failure.

PR #1942, opened 2026-07-06
chore: promote staging to main

noahlitvin:
## Summary

Promote staging → main. Contains:

- **Newest-first vault chart loading** (#1941): `fetchVaultStats` resolves `totalCount` in one request, then walks remaining pages newest-to-oldest via synthesized offset cursors (4 concurrent), streaming progressive partials into the react-query cache so the /vaults chart paints recent history immediately. maxPages 50 → 500 with oldest-pages-dropped semantics; interval refetches pull only the tail via a cached baseline.
- **Time-proportional chart x-axis**: numeric axis domain replaces the categorical default, so mixed snapshot cadences (3h history vs hourly) no longer stretch the recent window.
- **Robinhood 7/1 chart anchor**: vault charts on Robinhood/Meridian chains clamp visible history to the 2026-07-01 UTC launch and re-base % returns at the first visible point.
- **Positions "To Win" column**: sortable payout − size column on the positions table; settled/sold/closed rows show a dash and sink below open positions.

### DIFF
--- a/packages/app/src/components/positions/PositionsTable.tsx
+++ b/packages/app/src/components/positions/PositionsTable.tsx
@@ -345,13 +345,6 @@ function PositionRow({
           className="text-brand-white font-mono"
         />
       </TableCell>
-      <TableCell>
-        <NumberDisplay
-          value={payoutFormatted}
-          appendedText={collateralSymbol}
-          className="text-brand-white font-mono"
-        />
-      </TableCell>
       {/* To Win — net winnings (payout − size) if the position hits.
           Settled/sold/closed rows have nothing left to win. */}
       <TableCell>
@@ -364,6 +357,13 @@ function PositionRow({
         ) : (
           <span className="text-muted-foreground">—</span>
         )}
+      </TableCell>
+      <TableCell>
+        <NumberDisplay
+          value={payoutFormatted}
+          appendedText={collateralSymbol}
+          className="text-brand-white font-mono"
+        />
       </TableCell>
       {/* Profit/Loss → PENDING / CLAIM / Realized PnL */}
       <TableCell>{renderPnlCell()}</TableCell>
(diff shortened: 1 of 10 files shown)


noahlitvin:
## Review summary (multi-angle review, 9 verified findings → 7 survived)

**Fixed in 824997efc:**
1. ~~`fetchVaultStats` baseline refetch equated `baseline.length` with a server row offset~~ — a mid-walk failure could cache a newest-first partial that the retry treated as a `[0, N)` prefix, permanently truncating the series. Tail refetch now addresses the tail by `filter.timestamp.gte` (offset-independent), and `useVaultStats` only baselines from a **completed** walk (ref, not the partial-filled cache).
2. ~~Mid-walk failure rendered a truncated partial as complete, fresh data with no error indication~~ — the hook now rolls the cache back to the last complete series (or empty) on failure, and the chart shows an error message via a threaded `isError` prop.
3. ~~To Win ascending sort flooded "—" rows to the top~~ — resolved by simplification: To Win now always shows `payout − size` (both always exist), so the sentinel and the duplicated status classification are gone entirely.

**Known/accepted, fast-follow candidates:**
- Synthesized offset cursors couple the SDK to the API's cursor wire format (documented opaque). Verified byte-identical on both deployments today; a server-side cursor-scheme change would silently truncate to one page. Worth a server-side regression test or a first-class reverse-order arg.
- A `vault: null` response mid-walk is silently skipped (interior gap, no retry) — exotic trigger, defensive gap only.
- Per-batch full re-sort + one `setQueryData` re-render per 4-page batch on capped deployments; and `onProgress?.(series()); return series()` double-computes. Optimization opportunity, not correctness.

**Refuted during verification:** the Robinhood testnet anchor conflation (intentional, test-pinned) and stale-walk cache stomping (no reachable trigger; react-query dedupes in-flight fetches for this key).

